PR #2, opened 2017-03-23
Adding copy command to Move the binary to .helm plugin folder

rajivmucheli:
Moving the binary to .helm plugin folder is a simple and straight forward step.

As a beginner i did not know, helm diff had to be executed in the respective repo, hence added a common error a new user may under "Troubleshooting section"

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,14 @@ E.g.
 curl -L $TARBALL_URL | tar -C $(helm home)/plugins -xzv
 ```
 
+Post Downloading the desired release, do note you will need to move the binary in the helm plugins folder :
+
+```
+cp -r ~/Downloads/diff $HOME/.helm/plugins/
+```
+
+
+
 ## Build
 
 Clone the repository into your `$GOPATH` and then build it.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,6 +30,14 @@ E.g.
 curl -L $TARBALL_URL | tar -C $(helm home)/plugins -xzv
 ```
 
+Post Downloading the desired release, do note you will need to move the binary in the helm plugins folder :
```
databus23: `curl -L $TARBALL_URL | tar -C $(helm home)/plugins -xzv` is unpacking the tarsal directly in your helm plugins folder so this could be a little confusing. It should be made clear that this is an alternative (manual) installation method.

rajivmucheli:
Hi,

I tried this command but not sure why it failed in my case, as the intent is same! i wouldnt resubmit the request.

lachie83:
@rajivmucheli @databus23 It's probably easier just to document using helm plugin manager installation in Helm >2.3.x

PR #12, opened 2017-08-23
Enabling selecting filtering of diff output

### LINKED ISSUE #11: Allow suppression of actual diff by `kind`
First of all, super helpful tool that we use quite often!  Thanks for your time.

We would love to integrate `helm-diff` a bit tighter in our CI/CD pipeline but have a bit of an issue where we don't want to expose the actual `diff` of a credential to the output. 

It would be extremely handy to be able to blacklist a set of `kind` resources that would only show a `CONTENT CHANGED` type message if anything internal had changed instead of actually displaying the contents.

jrnt30:
Fixes #11 

- Allows easy suppression of secret content diff via the `-q` or `--suppress-secrets` flags 
- Allows additional customization of suppression via the `--suppress` flag

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,17 +22,15 @@ perform.
 var Version string = "HEAD"
 
 type diffCmd struct {
-	release string
-	chart   string
-	//	out     io.Writer
-	client helm.Interface
-	//	version int32
-	valueFiles valueFiles
-	values     []string
+	release         string
+	chart           string
+	client          helm.Interface
+	valueFiles      valueFiles
+	values          []string
+	suppressedKinds []string
 }
 
 func main() {
-
 	diff := diffCmd{}
 
 	cmd := &cobra.Command{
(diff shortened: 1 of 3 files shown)


databus23:
Thanks for the PR. I was on vacation in the last weeks and I"m still in the progress of catching up. I will try to look into this tomorrow. Thanks for your patience.

jrnt30:
Not a problem at all, hope you enjoyed your time off!

main.go
```diff
@@ -59,8 +62,10 @@ func main() {
 
 	f := cmd.Flags()
 	f.BoolP("version", "v", false, "show version")
+	f.BoolP("suppress-secrets", "q", false,  "suppress secrets in the output")
 	f.VarP(&diff.valueFiles, "values", "f", "specify values in a YAML file (can specify multiple)")
 	f.StringArrayVar(&diff.values, "set", []string{}, "set values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)")
+	f.StringArrayVar(&diff.suppressedKinds, "suppress", []string{}, "suppress secrets in the output")
```
databus23: The description is not accurate I think. It should mention You can specify kinds and that you have to specify it multiple times to exclude more then one.

databus23:
Thanks!

PR #18, opened 2018-01-11
FLAG: add --no-color to disable color in the output, fixes #17

### LINKED ISSUE #17: optional flag to disable color
There's a few use cases for removing the color from the diff, but the one of interest to me, is that the output without color can be added to github, and ` ```diff` will provide the proper syntax highlighting

jeff-knurek:


databus23:
Thanks!

jeff-knurek:
just noticed that I didn't update the README - not sure though if this feature needs to be documented there _(most other flags are missing from the README too)_

databus23:
@jeff-knurek You are right. I will take care eventually. :) Thanks

PR #20, opened 2018-01-19
Fix access to remote Helm repostories

elemental-lf:
I've had the problem that helm-diff failed when I specified a remote Helm chart on the command line. Investigation showed that ChartDownloader failed because it knew nothing about the http and https schemes. This PR fixes this problem. I'm using Helm version 2.7.2.

* Call ChartDownloader with a list of all available getters
* Return error from downloadTo to get better feedback why
  something went wrong

### DIFF
--- a/helm.go
+++ b/helm.go
@@ -13,6 +13,8 @@ import (
 
 	"github.com/ghodss/yaml"
 	"google.golang.org/grpc"
+	"k8s.io/helm/pkg/getter"
+	"k8s.io/helm/pkg/helm/environment"
 	"k8s.io/helm/pkg/downloader"
 	"k8s.io/helm/pkg/helm/helmpath"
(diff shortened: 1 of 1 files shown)


databus23:
Thank you!

aananthraj:
Hi,

Access to remote Helm repositories is great. 

Notised an issues: 
It downloads the remote chart and places in current directory. Downloading chart to Archive dir (~/.helm/cache/archive) would keep the current dir clean

Fix: #30

elemental-lf:
@aananthraj yeah, I also noticed this but it's not directly related to this PR.

aananthraj:
yes @elemental-lf , its no way related to this PR, but its a kind of cleanup around it. 
Just wanted to convey it here 😄 .

PR #29, opened 2018-02-16
Allow to omit large common parts in diff output

njam:
By default disabled, can be enabled and configured with `--context NUM`.
image

Wdyt?
This is my first time writing Go code, please let me know if my naming is off, something should be refactored or just be done differently!

The code iterates over all diff records and keep track of how many lines are between the closest change (addition or removal) *before* the current line. Then it iterates again through all diff records but backwards and adjusts the *distance* if the closest change *after* the current line is closer.
When printing the diff we can then simply omit records that have a *distance* larger than the desired *context*.

I have an alternative implementation which first creates *chunks* and then iterates over chunks of diffs to print, if you prefer it: [diff-with-chunks.go](https://gist.github.com/njam/5885e45f2a0590542f5a96a6391b8e56).

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,3 +182,11 @@ The above will install this plugin into your `$HELM_HOME/plugins` directory.
 - You need to have [Go](http://golang.org) installed. Make sure to set `$GOPATH`
 - If you don't have [Glide](http://glide.sh) installed, this will install it into
   `$GOPATH/bin` for you.
+
+### Running Tests
+Automated tests are implemented with [*testing*](https://golang.org/pkg/testing/).
+
+To run all tests:
+```
+go test -v ./...
+```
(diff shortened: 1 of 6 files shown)


njam:
@databus23 I've updated the branch to the new CLI interace.
Maybe a flag like `--context` which is not available in `helm upgrade` should go *before* the "upgrade" now? So it's `helm diff --context 3 upgrade ...`?

mumoshu:
@njam Thank you so much for contributing the awesome feature!

@zytek updated your work in #53 and I added my comments there. Would you mind taking a look?

njam:
@mumoshu I've added the `--context` option to all CLI commands that print a diff, and added a unit test.
wdyt?

njam:
@mumoshu when you have a moment, would you be able to review this?

njam:
@mumoshu let me know if something should be changed! :)

njam:
Cool, thanks!

PR #24, opened 2018-01-29
Feature Request: Diff to compare rollback

aananthraj:
This provides the ability to visualise what a helm rollback could change.
Usage:

```
$ helm diff rollback -h
This command compares the laset manifests details of a named release 
with specific revision values to rollback.

It forecasts/visualizes changes, that a helm rollback could perform.

Usage:
  diff rollback [flags] [RELEASE] REVISION

```

PR #22, opened 2018-01-24
feature Implementation: Diff between Revisions #6

aananthraj:
This provides the ability to do a diff on revisions.
I have made some improvements with 
- safer code terminations with (exit 0),
- Intelligent suggestions for subcommands
- docs
- code restructuring.
kindly have a look, and let me know in case of corrections/bugs.

Thanks..!!

PR #30, opened 2018-02-20
Download remote chart to Archive dir

aananthraj:
Downloading remote chart to Archive dir (~/.helm/cache/archive) would keep the current dir clean.

### DIFF
--- a/helm.go
+++ b/helm.go
@@ -76,7 +76,7 @@ func locateChartPath(name, version string, verify bool, keyring string) (string,
 		dl.Verify = downloader.VerifyAlways
 	}
 
-	filename, _, err := dl.DownloadTo(name, version, ".")
+	filename, _, err := dl.DownloadTo(name, version, helmpath.Home(homePath()).Archive())
 	if err == nil {
 		lname, err := filepath.Abs(filename)
 		if err != nil {


elemental-lf:
Sounds like a good idea to prevent clutter. Could there be any issues when running multiple helm instances in parallel? I haven't looked into DownloadTo to see if it does any locking or so.

aananthraj:
I just made it to mimics the way, Helm works. It works fine to me (as am using them often)
BTW what do you mean by running multiple helm instances in parallel? could you be brief?

elemental-lf:
@aananthraj I mean multiple instances of helm running under the same user account at the same time and so using the same archive directory. I haven't looked into the Helm source code if it does anything to prevent races between different processes when both try to download the same chart to the archive directory at the same time and so possibly corrupting the file or leading to failed verifications of the chart. If locking is done in downloadTo we'd be fine. If this isn't the case then we'd need to mimic the locking logic in helm diff, too, when writing to the shared archive directory.

aananthraj:
@elemental-lf , That was nice. Thanks..!!.

I have a query.
> when both try to download the same chart to the archive directory at the same time and so possibly corrupting the file or leading to failed verifications of the chart.

This could happen in PWD as well right..?, I mean when charts are downloaded to current working directory.

upon inspection codes of helm and helm diff plugin, I would like to share the following infrences.

helm diff plugin uses `DownloadTo` func from Helm's official `k8s.io/helm/pkg/downloader` package. 
>  If this isn't the case then we'd need to mimic the locking logic in helm diff, too, when writing to the shared archive directory

so logic is exactly the same.

elemental-lf:
This would also happen with the download to the cwd. But if there is a locking problem (at all), then it would be confined. I took a quick look at Helm now and I don't see any looking, so we're probably fine (or not worse than Helm itself).

aananthraj:
ya, exactly 😄 , can this be merged..? now.

databus23:
Thanks!

aananthraj:
Thanks @databus23  and @elemental-lf .
It would be helpful , If we could work together on PR #22 and #24 as well.

PR #33, opened 2018-04-03
Ensure values files exist

foklepoint:
Ensures that a value file actually exists instead of throwing up a golang exception when trying to open a file that doesn't exist


foklepoint:
@databus23 Can we please get this merged and cut a new version?

sstarcher:
@databus23 Pinged you on slack I understand if you don't have time, but would love to see this get merged.

databus23:
Thanks a lot!

PR #3, opened 2017-04-11
feat(core): Add ability to specify chart version in diff

lachie83:
Closes #1 

This is useful if you don't always want to use the latest chart in your diff. Allows user ability to specify chart version to diff against.

Changes
* fix usage typo
* add optional chart version to diff

Issue #1, opened 2017-03-21
(feature) - Allow chart version to diff to be specified


lachie83:

databus23:
I thought about this as well. Ideally it should be possible to compare arbitrary versions of a release.

I'm just not sure about the CLI UI for that without making it confusing. When comparing two *deployed* versions the command does not need the otherwise mandatory local directory.

lachie83:
Currently doing something like this as a workaround

```shell
helm fetch --untar deis/workflow --version $(WORKFLOW_VERSION)
helm diff deis-workflow ./workflow \
-f ../../env/$(CLUSTER_ID)/workflow-values.yaml
```

lachie83:
It also doesn't appear to trigger the hooks and display hook changes. Any ideas how hard that would be?

lachie83:
@databus23 I've submitted a PR with the ability to specify the chart version. LMK what you think

databus23:
I believe this has been fixed by a536dfc and is shipped since version 2.9.0+1 of this plugin.

databus23:
I believe this has been fixed by a536dfc and is shipped since version 2.9.0+1 of this plugin.

PR #34, opened 2018-04-03
Allow chart versioning

foklepoint:
We rename the existing version flag
and re-use the existing helm version flag.
This allow helm-diff to seamlessly work with
other helm tooling such as helmfile

### DIFF
--- a/main.go
+++ b/main.go
@@ -26,6 +26,7 @@ var Version = "HEAD"
 type diffCmd struct {
 	release         string
 	chart           string
+	chartVersion    string
 	client          helm.Interface
 	valueFiles      valueFiles
 	values          []string
@@ -42,8 +43,8 @@ func main() {
 		Short: "Show manifest differences",
 		Long:  globalUsage,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if v, _ := cmd.Flags().GetBool("version"); v {
-				fmt.Println(Version)
+			if v, _ := cmd.Flags().GetBool("diff-version"); v {
+					fmt.Println("Version", Version)
 				return nil
 			}
 
@@ -69,9 +70,10 @@ func main() {
 	}
 
 	f := cmd.Flags()
-	f.BoolP("version", "v", false, "show version")
+	f.BoolP("diff-version", "", false, "show diff version")
 	f.BoolP("suppress-secrets", "q", false, "suppress secrets in the output")
 	f.Bool("no-color", false, "remove colors from the output")
+	f.StringVarP(&diff.chartVersion, "version", "v", "", "specify the chart version")
 	f.VarP(&diff.valueFiles, "values", "f", "specify values in a YAML file (can specify multiple)")
 	f.StringArrayVar(&diff.values, "set", []string{}, "set values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)")
 	f.BoolVar(&diff.reuseValues, "reuse-values", false, "reuse the last release's values and merge in any new values")
@@ -84,7 +86,7 @@ func main() {
 }
 
 func (d *diffCmd) run() error {
-	chartPath, err := locateChartPath(d.chart, "", false, "")
+	chartPath, err := locateChartPath(d.chart, d.chartVersion, false, "")
 	if err != nil {
 		return err
 	}


foklepoint:
@databus23 can we get this merged? thoughts? We've been using this in production at reddit for a while now and it works fairly well.

sstarcher:
@databus23 this is a ticket that the https://github.com/roboll/helmfile project is waiting on

databus23:
This looks like a duplicate of #1

sstarcher:
@databus23 this is a PR that implements #1

databus23:
Sorry I mean't duplicate of #3

sstarcher:
Looks like #3 also implements this, but in a different way

sstarcher:
@databus23 I would think the `@version` would be more confusing than `--diff-version version`

databus23:
I'm fine with adding this feature in general. I don't like repurposing existing flags. Does it have to me `-v` or was that just so that it looks nice?

What I like about the `@version` is that it is part of the chart name. Specifying a flag `--chart-version` when `[CHART]` happens to be a local directory doesn't make sense.

sstarcher:
I don't think using `-v or --version` would make sense as that tends to imply the version of the plugin itself or verbose.

So from reading the other issues it looks like you have 2 requests.  
* Diffs between versions
* Diffs between the current release and a new chart

Current syntax `helm diff RELEASE CHART`.  If I'm following the code correctly CHART can be a local chart or a chart in a helm repository.  The version syntax makes no sense for a local chart.

If you are willing to break the current syntax I think it would make sense to differentiate these things.

`helm diff RELEASE CHART` for a local chart when CHART exists as a local file
`helm diff RELEASE@REVISION-REVISION2` for comparing 2 revisions
`helm diff RELEASE CHART@VERSION` for comparing the current release to a versioned chart

vs

`helm diff RELEASE CHART`
`helm diff RELEASE -r REVISION-REVISION2`
`helm diff RELEASE CHART -d VERSION`

Either syntax should work, but I think the second syntax would be easier to deal with in code.

benlangfeld:
What would the long-form versions of your proposed `-r` and `-d` options be, @sstarcher ?

sstarcher:
`--release` and `--diff-version` I'm fine with different names.  We could also go with `--chart-version` and use `-c`

benlangfeld:
I'm happy with that, but you guys know best.

sstarcher:
@benlangfeld ya, just trying to have the discussion so we can get some PR in to resolve this issue as this feature is much needed for anyone using helmfile.

sstarcher:
@foklepoint I would be fine with that.  If that's the case we should have a `helm diff version` command that display the version of the application.  Which would align us with what `helm version` does and `helm s3 version` also.

If implemented that way it would leave the diffing between releases to some other mechanism.

@databus23 opinions

foklepoint:
Yep that’s fair. I originally tried going that way but couldn’t get the flag/subcommand parsing to work easily so gave up :x

sstarcher:
So if we keep @foklepoint direction we could have

```
helm diff RELEASE CHART # Diff the latest chart version
helm diff RELEASE -r REVISION-REVISION2 # Diff between releases
helm diff RELEASE CHART --version VERSION # Diff a new chart version
```

mumoshu:
@sstarcher Looks awesome to me. And a `helm diff version` sub-command to replace the existing `helm diff --version` functionality in a same PR?

mumoshu:
@databus23 Do you worry about the backward compatibility of the `--version` flag? 😉 
If so, we still need to named it e.g. `--chart-version` as you've suggested.

So, it seems like:

```
helm diff RELEASE CHART # Diff the latest chart version
helm diff RELEASE -r REVISION-REVISION2 # Diff between releases
helm diff RELEASE CHART --version VERSION # Diff a new chart version
helm diff version # Show the version number
```

vs

```
helm diff RELEASE CHART # Diff the latest chart version
helm diff RELEASE -r REVISION-REVISION2 # Diff between releases
helm diff RELEASE CHART --chart-version VERSION # Diff a new chart version
helm diff --version # Show the version number
```

databus23:
Ok, after some thought I'm now actually for using `--version` to specify the version of a chart.
The main reason for this being that this is how `helm upgrade` works and its nice to keep `helm diff` *flag compatible*.
It allows to "prepare" an upgrade using helm diff und when the output of `helm diff` looks good all that is left to do is to change `helm diff` to `helm upgrade` to apply the changes. At least this is how I have used helm diff in the past when performing upgrades manually.

In general I would propose the following changes.

* Remove `-v` and add a `helm diff version` subcommand (same way helm works)
* Add `--version` to specify the chart version
* Add a `helm diff chart RELEASE CHART` subcommand
* Alias `helm diff` to `helm diff chart`
* (Not in this PR: Add `helm diff revisions RELEASE [FROM] [TO]` command)

I think without adding subcommands it will be difficult to add support for diffing revisions which has been requested a few times by now. Aliasing the root command to `helm diff chart` should provide backward compatibility (unless your release is called "version" or "revisions" 😄  )

Any comments on that?

sstarcher:
@databus23 I think that all makes great sense. @foklepoint would you be able to implement the `helm diff chart` and `helm diff version` in this?  I imagine we can leave `helm diff revisions` to a separate PR.

databus23:
I'm offering to restructure the cli code to add the subcommands and alias (it requires bumping cobra and pflag to get https://github.com/spf13/cobra/issues/295) if the majority is fine with the proposal.

sstarcher:
@databus23 I'm 100% on board with your proposal I think it makes the most sense.

foklepoint:
@databus23 Fine with your proposal, feel free to take over this PR if you need to. I don't have enough cycles to address the changes proposed here in the short term, but we'd definitely love to start using upstream helm-diff again :)

databus23:
I pushed the proposed changes to master (`helm diff chart` became `helm diff upgrade`).
I also added the `--version` flag to `diff upgrade`.

Would anybody care to test this and give some feedback if they see any problem with it, especially in regards to backward compatibility?

sstarcher:
@databus23 looks like `helm diff` is broken
```helm diff dex stable/dex --version 0.1.0                                                                                                                 
Command "helm diff" is deprecated, use "helm diff upgrade" instead```

**Comment by sstarcher** [comment]
Tested out `helm diff upgrade prometheus stable/prometheus --version 6.2.0` and that worked great.

**Comment by sstarcher** [comment]
also `helm diff version` looks good

**Comment by databus23** [comment]
@sstarcher That is expected. I added this encourage people to the the new subcommand but the command should still work. Is the command exiting cleanly? Then you maybe just have an empty diff.

**Comment by databus23** [comment]
So to clarify: `helm diff` should work exactly like `helm upgrade diff` only that it shows a deprecation notice.

**Comment by sstarcher** [comment]
@databus23 yep, my mistake I assumed it existed with a error, but you are correct.  It all LGTM.

**Comment by sstarcher** [comment]
@databus23 ahh the issue I saw was a different issue.  Looks like helm diff does not work work helm s3

```
Command "helm diff" is deprecated, use "helm diff upgrade" instead
Error: Get s3://CHART/charts/CHART-0.1.1.tgz: unsupported protocol scheme "s3"
Usage:
  diff [flags]
  diff [command]

Available Commands:
  upgrade     Show a diff explaining what a helm upgrade would change.
  version     Show version of the helm diff plugin

Flags:
  -h, --help                   help for diff
      --no-color               remove colors from the output
      --reset-values           reset the values to the ones built into the chart and merge in any new values
      --reuse-values           reuse the last release's values and merge in any new values
      --set stringArray        set values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)
      --suppress stringArray   allows suppression of the values listed in the diff output
  -q, --suppress-secrets       suppress secrets in the output
  -f, --values valueFiles      specify values in a YAML file (can specify multiple) (default [])
      --version string         specify the exact chart version to use. If this is not specified, the latest version is used

Additional help topics:
  diff

Use "diff [command] --help" for more information about a command.

```

I'll open a separate ticket

databus23:
I believe this has been fixed by a536dfc and is shipped since version 2.9.0+1 of this plugin.

PR #36, opened 2018-04-11
FLAG: add an option to ignore the error if the release doesn't exist yet

jeff-knurek:
I ran into the issue described here today https://github.com/databus23/helm-diff/pull/15 
I kind of agree that `helm-diff` shouldn't be responsible for replacing `--dry-run --debug`/`template`

but when building `helm diff` into CI, a new project might not always exist, and suppressing all errors from `helm diff` is not desirable. So I created this simplified version of the original PR.

### DIFF
--- a/main.go
+++ b/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/mgutz/ansi"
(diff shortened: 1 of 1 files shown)


jeff-knurek:
hmm, that's a good point. I guess that's why I only "_kind of agreed_" 😄

Issue #6, opened 2017-04-13
Diff between Revisions


sfrique:
Hello,

It would be great if it was possible to diff between already released versions.

Something like this, current release history:
```
$ helm history winsome-owl                                                                                                                                             [11:04:28]
REVISION        UPDATED                         STATUS          CHART           DESCRIPTION     
1               Wed Apr 12 18:23:28 2017        SUPERSEDED      api-0.1.0    Install complete
2               Wed Apr 12 18:25:55 2017        SUPERSEDED      api-0.1.0    Upgrade complete
3               Wed Apr 12 18:30:48 2017        SUPERSEDED      api-0.1.0    Upgrade complete
4               Wed Apr 12 18:32:39 2017        SUPERSEDED      api-0.1.0    Rollback to 2   
5               Thu Apr 13 00:08:00 2017        DEPLOYED        api-0.1.0    Upgrade complete
```
So to get a diff between revision 5 (current and 3)

```
$ helm diff winsome-owl 3
```
Or explicit get the diff between 4 and 1
```
$ helm diff winsome-owl 4-1
```

Thanks

aananthraj:
Waiting for this feature..!!
Mean while, as a work around , we could use the `helm get` and  `sdiff`/`vimdiff` to view the difference between two revisions side by side.

**Using sdiff**:
```bash
$ sdiff <(helm get $release --revision $revision_1 ) <(helm get $release --revision $revision_2)
 ```
To display only the changes use `-s` and `-w` flags
```bash
$ sdiff -s -w 80 <(helm get $release --revision $revision_1 ) <(helm get $release --revision $revision_2)
 ```

**Using vimdiff**:
```bash
$ vimdiff <(helm get $release --revision $revision_1 ) <(helm get $release --revision $revision_2)
 ```
Example:
```bash
$ sdiff <(helm get my_release --revision 4 ) <(helm get my_release --revision 5 )
REVISION: 4						      |	REVISION: 5
RELEASED: Fri Dec 15 08:59:48 2017			      |	RELEASED: Fri Dec 15 09:02:09 2017
...                                                           | ....
```

aananthraj:
hi @sfrique,
I have implemented the above mentioned feature in this PR https://github.com/databus23/helm-diff/pull/21, have a look and let me know in case of corrections/bugs.

sfrique:
@aananthraj  Thanks a lot to take sometime to implement this!

Unfortunately I don't have the time to test it now =/
[...]

PR #39, opened 2018-04-12
FLAG: allow option for chart version

jeff-knurek:
The first issue #1 & one of the first PRs are still open on this topic.
I think #6 / #22 is an entirely unrelated proposal.

If `helm` allows for different version on install (with `--version`), `helm diff` should be able to determine the diff of what gets installed with that (with this PR using `--chartVersion`).

### DIFF
--- a/main.go
+++ b/main.go
@@ -26,6 +26,7 @@ var Version = "HEAD"
 type diffCmd struct {
 	release         string
 	chart           string
+	chartVersion    string
 	client          helm.Interface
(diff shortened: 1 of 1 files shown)


databus23:
Duplicate of #34. Lets continue there as already a lot of people commented there.

jeff-knurek:
yeah, sorry - it completely makes sense to close this one (I didn't see it before started) - I'm just shocked/disappointed that this issue hasn't been resolved yet

PR #41, opened 2018-04-13
add namespace to deduplication tuple

novas0x2a:
helm charts can deploy to multiple namespaces at once, so compare on
(namespace, name, kind, version) instead of just (name, kind, version).

(We're deploying ResourceQuotas with the same name across multiple namespaces)

### DIFF
--- a/cmd/revision.go
+++ b/cmd/revision.go
@@ -91,7 +91,12 @@ func (d *revision) differentiate() error {
 			return prettyError(err)
 		}
 
-		diff.DiffManifests(manifest.Parse(revisionResponse.Release.Manifest), manifest.Parse(releaseResponse.Release.Manifest), d.suppressedKinds, d.outputContext, os.Stdout)
+		diff.DiffManifests(
+			manifest.Parse(revisionResponse.Release.Manifest, revisionResponse.Release.Namespace),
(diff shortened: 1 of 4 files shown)


databus23:
Did you test this? In my case the Namespace is empty. I guess thats because it was not specified in the templates rendered by helm. Maybe we should provide a string like `RELEASE_NAMESPACE` for namespace instead of rendering it as empty.

databus23:
Or maybe we even pass in the release namespace into the `Parse` method to fill it in if its not specified in the spec.

novas0x2a:
Okay, finally got around to learning more about the helm api; this seems to do the right thing now.

novas0x2a:
Noticed there was a conflict; rebased it.

mumoshu:
@novas0x2a Sorry but this got to conflict badly with the current master. Would you mind rebasing when you have some time? I'm still considering this as an important edge-case to be addressed.

novas0x2a:
rebased again.

novas0x2a:
@mumoshu @databus23?

mumoshu:
@novas0x2a Sorry it took some time to test this manually on my side. LGTM. Thanks for your patience, and awesome work!

PR #43, opened 2018-04-20
fix diff binary path in plugin.yaml

codenio:
Hi, 

The path specified in the plugin.yaml points to `~/.helm/plugins/helm-diff/bin/diff` while the binary is available at `~/.helm/plugins/helm-diff/diff`. This creates following error
```bash
$ helm diff -h
Error: fork/exec ~/.helm/plugins/helm-diff/bin/diff: no such file or directory
```

### DIFF
--- a/plugin.yaml
+++ b/plugin.yaml
@@ -5,6 +5,6 @@ version: "2.8.2+2"
 usage: "Preview helm upgrade changes as a diff"
 description: "Preview helm upgrade changes as a diff"
 useTunnel: true
-command: "$HELM_PLUGIN_DIR/bin/diff"
+command: "$HELM_PLUGIN_DIR/diff"
 hooks:
   install: "$HELM_PLUGIN_DIR/install-binary.sh"


databus23:
This is already fixed in master, no?

databus23:
See https://github.com/databus23/helm-diff/commit/380325c4249b20fcbbfedb55865729dfbb18d363

databus23:
Ah see, it’s still broken in the Makefile. I‘ll fix momentarily.

aananthraj:
can you be specific, where in Makefile..?

databus23:
Here https://github.com/databus23/helm-diff/commit/993b7db51cf45c486148791da0c251a3e28363ed

aananthraj:
is it intended to keep under `bin` (~/.helm/plugins/helm-diff/**bin**/diff ) folder..?

databus23:
The reason is there is now a `diff` folder in repository containing the diff related go code. So the binary had to move.

aananthraj:
yes that is with source code right..??
what about the destination..? I mean inside the plugin directory, is /bin/diff required..?

PR #44, opened 2018-04-20
fixes to be able to set tillernamespace for diff command

rmartinez3:
This pull request is so that you can specify a tiller instance located in another namespace other than kube-system. 

Add --tiller-namespace flag so can set the namespace of the tiller instance.

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -14,6 +14,7 @@ type diffCmd struct {
 	release         string
 	chart           string
 	chartVersion    string
+	tillerNamespace string
 	client          helm.Interface
 	valueFiles      valueFiles
 	values          []string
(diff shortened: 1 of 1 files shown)


rmartinez3:
helm itself has a flag --tiller-namespace , adding the flag to diff plugin alllows the --tiller-namespace flag to be used. Which then sets what tiller to tunnel to . I tried to see via the helm plugin library if it can be set to which tiller you want to use via namespace. However i noticed that just setting the flag is enough. The variable can be removed since its not used. 

However adding a flag will allow to specify tiller in different namespaces. 

Another way is to set $TILLER_NAMESPACE environment variable . However , I will prefer to use a flag to specify tiller namespace.

databus23:
This sounds to me like a bug in helm. It already processes other flags without the need to explicitly specify them in the plugin.
See: https://github.com/kubernetes/helm/blob/master/docs/plugins.md#a-note-on-flag-parsing

Could you check with the helm core people if this is just an oversight. Sure looks like it to me.

rmartinez3:
Yea , ill ping this ticket on the reply i get from them. Thanks for checking on this.

databus23:
Looking at the code in helm. Specifically this: https://github.com/kubernetes/helm/blob/4d519a741dd43f8f0c175228dee9e9df951a7c5c/cmd/helm/load_plugins.go#L124

It only works if the flag is passed without a space and using an `=`:

* `--tiller-namespace=kube-system` works
* `--tiller-namespace kube-system` doesn't work (`unknown flag: --tiller-namespace`)

rmartinez3:
ahh that works . Thanks for looking into this . will close pull request

databus23:
@rmartinez3 If opened a PR to fix this upstream: https://github.com/kubernetes/helm/pull/3944

rmartinez3:
Thanks saw the PR got merged. Now can set flag both ways when 2.9.0 is out .

Issue #50, opened 2018-04-27
helm plugin install does not pin versions


clux:
Hey,

I did a rebuild of an image today with pinned helm and helm diff plugin using:

```
helm plugin install https://github.com/databus23/helm-diff --version v2.8.2+2
```

and ended up getting a version of helm diff that now required an upgrade subcommand. Is this way of version pinning not meant to work, or is the semver tag actually considered floating?

databus23:
Hmm that should work. 2.8.2+2 doesn't have the cli refactoring in. what does `helm diff -v` say (or `or helm diff version` if that doesn't work)

Looks like you got v2.9.0+1

clux:
Yep, I totally got v2.9.0+1 (only last command worked).

CI logs:

```
+ helm plugin install https://github.com/databus23/helm-diff --version v2.8.2+2
*   Trying 192.30.253.116...
* TCP_NODELAY set
* Connected to api.github.com (192.30.253.116) port 443 (#0)
* ALPN, offering http/1.1
* successfully set certificate verify locations:
  CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: /etc/ssl/certs/
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
} [512 bytes data]
* TLSv1.2 (IN), TLS handshake, Server hello (2):
{ [108 bytes data]
* TLSv1.2 (IN), TLS handshake, Certificate (11):
{ [3091 bytes data]
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
{ [333 bytes data]
* TLSv1.2 (IN), TLS handshake, Server finished (14):
{ [4 bytes data]
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
} [70 bytes data]
* TLSv1.2 (OUT), TLS change cipher, Client hello (1):
} [1 bytes data]
* TLSv1.2 (OUT), TLS handshake, Finished (20):
} [16 bytes data]
* TLSv1.2 (IN), TLS change cipher, Client hello (1):
{ [1 bytes data]
* TLSv1.2 (IN), TLS handshake, Finished (20):
{ [16 bytes data]
* SSL connection using TLSv1.2 / ECDHE-RSA-AES128-GCM-SHA256
* ALPN, server accepted to use http/1.1
* Server certificate:
*  subject: C=US; ST=California; L=San Francisco; O=GitHub, Inc.; CN=*.github.com
[...]

PR #51, opened 2018-05-03
fix install-binary.sh version specification

kujenga:
This fixes a bug where the --version flag that can be passed to `helm
plugin install` was not working with the install-binary.sh script that
this repository uses to install the binary files for users.

The issue was that the Helm runs install hooks in the current working
directory [1], so the `git describe` command that this repository was
executing was not being executed in the directory of the plugin, and
would thus always fail. The fix here is to use the `-C` flag to git to
specify the directory of the plugin source, ensuring that it will work
correctly regardless of the working directory from which the script is
executed.

Closes https://github.com/databus23/helm-diff/issues/50

[1] https://github.com/kubernetes/helm/blob/401f8bcc187d7a9edeec12c26c64819f105954d5/cmd/helm/plugin.go#L48-L72

### DIFF
--- a/install-binary.sh
+++ b/install-binary.sh
@@ -67,7 +67,7 @@ verifySupported() {
 # getDownloadURL checks the latest available version.
 getDownloadURL() {
   local url="https://api.github.com/repos/$PROJECT_GH/releases/latest"
-  local version=$(git describe --tags --exact-match 2>/dev/null)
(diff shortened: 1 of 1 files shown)


databus23:
Thanks for figuring this out!

kujenga:
Sure thing!

PR #52, opened 2018-05-16
fix: make dist should put diff binaries in bin folder

hoesler:


mumoshu:
@hoesler Hi. Thanks for the PR!

Would you mind clarifying your original purpose of this fix?

I thought that you wanted to install the plugin via a release tarball i.e. `helm plugin install https://github.com/databus23/helm-diff/releases/download/v2.9.0%2B1/helm-diff-macos.tgz`.

But as far as I can see, your fix alone isn't enough for that. When I actually run the command, it fails like this:

```
$ helm plugin install https://github.com/databus23/helm-diff/releases/download/v2.9.0%2B1/helm-diff-macos.tgz
Error: plugin metadata (plugin.yaml) missing
```

Looking into the tarball, it seems like we need to flatten the contents of the tarball in order for `helm install` to notice the `plugin.yaml`?

```
$ tar tzvf helm-diff.tgz
drwxr-xr-x  0 fabian staff       0  4 27 07:58 diff/
-rw-r--r--  0 fabian staff   11358  4 27 07:58 diff/LICENSE
-rw-r--r--  0 fabian staff     328  4 27 07:58 diff/plugin.yaml
-rwxr-xr-x  0 fabian staff 15298900  4 27 07:58 diff/diff
-rw-r--r--  0 fabian staff     1510  4 27 07:58 diff/README.md
```

databus23:
Thanks for bringing this up.
The binary was moved to the `bin` folder because the normal installation procedure clones the repo and since the last refactor there now is a folder `diff`, hence the binary can't be placed in the plugin root anymore.

PR #53, opened 2018-05-23
Short diff output (#29 with updates)

zytek:
Just took #29 and updated it to work with master.

### DIFF
--- a/cmd/revision.go
+++ b/cmd/revision.go
@@ -17,6 +17,7 @@ type revision struct {
 	client          helm.Interface
 	suppressedKinds []string
 	revisions       []string
+	outputContext   int
 }
 
 const revisionCmdLongUsage = `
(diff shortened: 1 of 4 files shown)


cmd/revision.go
```diff
@@ -69,6 +70,7 @@ func revisionCmd() *cobra.Command {
 
 	revisionCmd.Flags().BoolP("suppress-secrets", "q", false, "suppress secrets in the output")
 	revisionCmd.Flags().StringArrayVar(&diff.suppressedKinds, "suppress", []string{}, "allows suppression of the values listed in the diff output")
+	revisionCmd.Flags().IntVarP(&diff.outputContext, "context", "C", -1, "output NUM lines of context around changes")
```
mumoshu: For anyone interested - this corresponds to the equivalent option in the `diff` command:

```console
$ diff --help | grep -- -C
  -c  -C NUM  --context[=NUM]  Output NUM (default 3) lines of copied context.
```

njam:
> adding a single test-case to printDiff

I will gladly do it, to learn how it's done! I can do it on the next weekend. If somebody else wants to do it beforehand, feel free.
Just to be sure, you mean to add a *unit test* that calls `printDiff()` and checks the return value right?

mumoshu:
@zytek @njam I've finally reviewed and merged #29. So closing this in favor of that. Thanks for your patience!

PR #63, opened 2018-06-20
Fix install command

KernelPanicAUS:


mumoshu:
@KernelPanicAUS Thanks for the PR! Just curious but what would be the point of explicitly specifying the version=master here? I was thinking that is the default behavior anyway.

KernelPanicAUS:
From my experience helm has some issues with git 2.16+ where it can't do a git checkout without specifying the branch.

Specifying the branch might appear superfluous, but it makes the barrier to entry lower for folks who are just getting into helm and its plugin ecosystem and saves them from having to research why the copy-pasted command didn't work :)

mumoshu:
@KernelPanicAUS Thanks for pointing that out! I slightly remember that I learned about that in the past. So, this would indeed help users like me 😆

Issue #54, opened 2018-05-25
Return exit code


royjs:
I am deploying with Helm from CI and it would be really helpful to have the commands return a different exit code based on whether there is some changes or not. From CI I usually don't care about the actual changes but I want to know if something changed or not.

My suggestion would be to have a `--detailed-exit-code` flag similar to `terraform plan` or `git diff`

The best would probably to do like terraform and return something like this:

- 0 if no error and no diff
- 1 if there is an error
- 2 if no error but a diff

This change would also allow to implement the same flag in `helmfile`. See roboll/helmfile#133

Any thoughts?

travisgroth:
I'd like to see this as well so that a helm deployment is only triggered if there is an actual difference.   The lack of information as to whether a deployment is needed has been our biggest issue with helm and integrating it into more 'driverless' automation.

The exit code is actually easy to implement in helm-diff based on my testing, but helm itself will treat any non-zero exit code as an error.  So, this should probably be coupled with a change in helm to define an exit code contract to signal failure versus a code that should be passed back to the caller.

royjs:
@travisgroth thanks for looking into this, I didn't have time to do it yet.
It looks like it is a little more complicated than what I thought. 
Were you planning on opening an issue in the helm project ? I can also do it if you prefer

travisgroth:
@royjs it probably makes sense for you to initiate the issues across the 3 projects, but I'll chime in on the discussion.   Assuming there is an agreement on the path forward and the interface, I may be able to do a PR for one or more of these since I've already wandered through the code path.

mumoshu:
@travisgroth @royjs @cmeury Would you be able to participate in the helm part of this issue at https://github.com/helm/helm/issues/4170?  As a workaround, I was trying to bypass helm's plugin system, hence call `helm-diff` directly. But turned out it isn't very feasible as explained in the issue. I'm now eager to implement this in helm. Appreciate your support there 😃

mumoshu:
https://github.com/helm/helm/pull/4367 was merged to the upstream to resolve https://github.com/helm/helm/issues/4170.

It should be good to start returning the separate exit code for no error but diff now.

mumoshu:
#78 in combination with `helm` built from the current master seems to work:

```
$ helm211dev diff upgrade myrelease stable/mysql --detailed-exitcode
...
$ echo $?
2
```

@royjs @travisgroth @cmeury Could you verify that it works for you?

[...]

PR #65, opened 2018-06-26
return non-zero exitcode when encountering any change on upgrade

cmeury:
An attempt at resolving #54, maybe someone can pick it up and build on it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ Examples:
 
 Flags:
   -h, --help                   help for upgrade
+      --detailed-exitcode      return a non-zero exit code when there are changes
       --reset-values           reset the values to the ones built into the chart and merge in any new values
       --reuse-values           reuse the last release's values and merge in any new values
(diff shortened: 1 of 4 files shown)


mumoshu:
@cmeury Thanks for your contribution across projects 👍This LGTM overall, would you mind rebasing?

Also, I'm fine with adding `os.Exit(2)` on no error but diff(s) in this PR if it's possible. `helm diff` would return 1 anyway due to the helm restriction, but I guess that we can avoid that by running helm-diff directly.

mumoshu:
Update:

> `helm diff` would return 1 anyway due to the helm restriction, but I guess that we can avoid that by running helm-diff directly.

This turned out to be not that easy(obviously? 😃) due to that helm-diff needs `TILLER_HOST` to be set to the tunneled tiller endpoint, which requires us to run helm-diff via the helm plugin system.

mumoshu:
Trying to rebase this onto master.

Issue #67, opened 2018-06-26
Install fails when running on IPs that are Rate Limited by GH API


helgi:
Offices and sometimes build boxes can get rate limited when using the unauthenticated API

helgi:
Closed by #69

PR #69, opened 2018-06-27
download from a known GH location when specifying version

helgi:
Latest still uses the GH API


helgi:
Closes #67 but only when people specify version, not on `latest`

Issue #35, opened 2018-04-11
TLS secured Tiller doesn't work


tboerger:
Hi,

my Helm installation is done with TLS, since the diff sub command doesn't provide any `--tls` flag I'm not able to use this tool with my setup :(

sprutner:
+1, I would love to see support for this.

maver1ck:
@mumoshu 
Any new release with that ?

mumoshu:
@maver1ck Just released v2.9.0+3!

PR #71, opened 2018-07-11
feat: TLS-based client auth support

vdesjardins:
This pull request add support for connecting to a Tiller server secured with TLS.


vdesjardins:
fixes #35

PR #72, opened 2018-07-12
feat: Capability to diff changes in chart hooks

rmartinez3:
Currently helm diff does not parse helm hooks. This appends hooks to the manifest diffs so that it can include them to be diff. Added capability to revision,upgrade and rollback.


mumoshu:
@MattKetmo Hey! That's a good point. Would you mind submitting a PR for that?

PR #73, opened 2018-07-13
feat: Improved diff on apiVersion changes

### LINKED ISSUE #70: moving between apiVersions produces a full remove/add
Use case is moving a Deployment or DaemonSet to `apps/v1` - Helm diff shows it as a full remove and add but helm/k8s seem to treat it like a bump. Not seeing the `Age` or anything like that change.

novas0x2a:
Resolves #70

When the apiVersion of a Kind changes, helm-diff detects this as a
remove/readd of the whole manifest; this makes it difficult to see the
actual changes. This change enables a small heuristic so apiVersion
changes like `authentication.k8s.io/v1beta1` ->
`authentication.k8s.io/v1` show up as a one-line diff instead of a
full-manifest diff, while hopefully not meddling with unexpected
apiVersions too badly.

### DIFF
--- a/manifest/parse.go
+++ b/manifest/parse.go
@@ -28,7 +28,13 @@ type metadata struct {
 }
 
 func (m metadata) String() string {
-	return fmt.Sprintf("%s, %s, %s (%s)", m.Metadata.Namespace, m.Metadata.Name, m.Kind, m.ApiVersion)
+	apiBase := m.ApiVersion
+	sp := strings.Split(apiBase, "/")
+	if len(sp) > 1 {
+		apiBase = strings.Join(sp[:len(sp)-1], "/")
+	}
+
+	return fmt.Sprintf("%s, %s, %s (%s)", m.Metadata.Namespace, m.Metadata.Name, m.Kind, apiBase)
 }
 
 func scanYamlSpecs(data []byte, atEOF bool) (advance int, token []byte, err error) {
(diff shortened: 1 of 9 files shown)


helgi:
@mumoshu @databus23 Would be useful to have this in a release as people are starting to move from older APIs (beta) to the stable ones for Deployments and such; and it kinda freaks them out seeing an add/remove kind of diff when that isn't the case :)

manifest/parse.go
```diff
@@ -28,7 +28,13 @@ type metadata struct {
...
 
 func (m metadata) String() string {
-	return fmt.Sprintf("%s, %s, %s (%s)", m.Metadata.Namespace, m.Metadata.Name, m.Kind, m.ApiVersion)
+	apiBase := m.ApiVersion
+	sp := strings.Split(apiBase, "/")
+	if len(sp) > 1 {
+		apiBase = strings.Join(sp[:len(sp)-1], "/")
+	}
+
+	return fmt.Sprintf("%s, %s, %s (%s)", m.Metadata.Namespace, m.Metadata.Name, m.Kind, apiBase)
```
mumoshu: For anyone interested - This is the heuristic to treat both apiVersions `apps/v1` and `apps/v1beta` as just `apps`, so that we can [give the same manifest name in mapping results](https://github.com/databus23/helm-diff/blob/master/manifest/parse.go#L90-L99) for appropriate diffing.

PR #78, opened 2018-07-25
feat: Dedicated exit code (2) on there's diff

### LINKED ISSUE #54: Return exit code

mumoshu:
This feature is enabled only when the `--detailed-exitcode` flag (#65) is provided.
Supports `upgrade`, `revision` and `rollback` sub-commands.

Resolves #54, but only after https://github.com/helm/helm/pull/4367 is released with helm v2.11.0, probably.


mumoshu:
Merging this to start gathering more feedbacks on #54

PR #82, opened 2018-08-21
Bump Helm to v2.10.0

### LINKED ISSUE #81: Helm-diff for helm 2.10
As we have already helm v2.10-rc.2 could we also have helm-diff plugin for it ?

Using 2.9.0 with helm 2.10 gives error:
Error: incompatible versions client[v2.9.0] server[v2.10.0-rc.2]

mumoshu:
Ran `glide update` and `glide install --strip-vendor`, followed by manual changes to `glide.lock` to revert `k8s.io/*` pkg versions to those seen in helm's glide.lock https://github.com/helm/helm/blob/cc7cc1087f586f64d0b0d08aa516a003ced2d85b/glide.lock#L457 to avoid build errors.

Resolves #81

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,10 +1,15 @@
-hash: 577ed63de6e9ee86cafc0e6051f2aaa2095ec5c3f009fa5d986b652c0f041495
-updated: 2018-07-12T17:35:13.372814615-07:00
+hash: bcc7754b1b92cc5ed4d3cb9ae50d7b343803e058fde4ab232973ca3b2f0e99eb
+updated: 2018-08-27T11:16:21.783008205+09:00
 imports:
 - name: github.com/aryann/difflib
   version: a1a4dd44eb11820695fbe83e00fb2301ee6eb54c
 - name: github.com/BurntSushi/toml
   version: b26d9c308763d68093482582cea63d69be07a0f0
+- name: github.com/docker/distribution
+  version: edc3ab29cdff8694dd6feb85cfeb4b5f1b38ed9c
+  subpackages:
+  - digestset
+  - reference
 - name: github.com/ghodss/yaml
   version: 73d445a93680fa1a78ae23a5839bad48f32ba1ee
 - name: github.com/gobwas/glob
@@ -17,6 +22,13 @@ imports:
   - syntax/lexer
   - util/runes
   - util/strings
+- name: github.com/gogo/protobuf
+  version: c0656edd0d9eab7c66d1eb0c568f9039345796f7
+  subpackages:
+  - proto
+  - sortkeys
+- name: github.com/golang/glog
+  version: 44145f04b68cf362d9c4df2182967c2275eaefed
 - name: github.com/golang/protobuf
   version: 1643683e1b54a9e88ad26d98f81400c8c9d9f4f9
   subpackages:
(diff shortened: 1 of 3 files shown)


glide.yaml
```diff
@@ -5,8 +5,17 @@ import:
...
+  version: v2.10.0
+  subpackages:
+  - pkg/downloader
+  - pkg/getter
+  - pkg/helm/environment
+  - pkg/helm/helmpath
+  - pkg/nstrvals
+  - pkg/helm
+  - pkg/tlsutil
+  - pkg/proto/hapi/release
```
mumoshu: I'm not entirely sure if this is good thing to do, but tried my best not to install unnecessary subpackages to vendor/ while building the project.

glide.lock
```diff
@@ -96,20 +119,110 @@ imports:
...
+  - storage/v1
+  - storage/v1alpha1
+  - storage/v1beta1
+- name: k8s.io/apiextensions-apiserver
+  version: 898b0eda132e1aeac43a459785144ee4bf9b0a2e
+  subpackages:
+  - pkg/features
 - name: k8s.io/apimachinery
-  version: 01bc873149a1802eb74df583613872d126449ed5
+  version: f6313580a4d36c7c74a3d845dda6e116642c4f90
```
mumoshu: After running `glide update`, I've manually updated this to the version that is also seen in [helm's glide.lock](https://github.com/helm/helm/blob/cc7cc1087f586f64d0b0d08aa516a003ced2d85b/glide.lock#L401).
From my observation, it was required in order to fix build errors due to version mismatch between apimachinery and [json-iterator](https://github.com/databus23/helm-diff/pull/82/files#diff-f16a80eae23d5b298c2652448ec420cfR45).

glide.lock
```diff
@@ -126,11 +240,87 @@ imports:
   - pkg/provenance
   - pkg/repo
   - pkg/resolver
+  - pkg/storage/driver
   - pkg/strvals
   - pkg/sympath
   - pkg/tlsutil
   - pkg/urlutil
   - pkg/version
+- name: k8s.io/kubernetes
```
mumoshu: Dislike the huge additional dependency here, but `glide update` added it anyway.

databus23: Urks. How did the helm people managed to bring that back in... :(

databus23: Ah man. It seems the only reason k8s.io/kubernetes is pulled in because of `fake.go` in k8s.io/helm/pkg/helm. And only because they use 3 errors from the `k8s.io/helm/pkg/storage` in there. This is so unnecessary...

databus23: This is clearly an accident. We should try to fix this upstream. We have had this before. https://github.com/helm/helm/pull/2685

databus23: So just by replacing `storage.ErrReleaseNotFound(rlsName)` with `fmt.Errorf("release: %q not found", rlsName)` three times in `fake.go` the kubernetes dependency goes away from `k8s.io/pkg/helm`

databus23: I'm opening a PR upstream

glide.lock
mumoshu: According to the discussion on the k8s slack, I'm going to add respective `version: ...` settings to affected dependencies, so that we won't need to manually update these. Thx for your review @databus23!

mumoshu:
@databus23 Awesome! And the plan looks great. Thanks for your support.

mumoshu:
So I'm going to merge this to resolve #81 for now.
Hoping the upstream PR is reviewed relatively soon, so that we can fork and alter the dependency to point the fork for smaller binary size.

Issue #83, opened 2018-08-21
--set flag parses numbers as float64


kujenga:
This is effectively the same issue as one filed in Helm upstream, https://github.com/helm/helm/issues/1707, which was resolved in this PR, which adds a `--set-string` flag to set string values specifically: https://github.com/helm/helm/pull/3599

In order to handle this error case, I believe that helm-diff needs to add support for this flag as well.

kujenga:
PR is now open for this here: https://github.com/databus23/helm-diff/pull/84

mumoshu:
@kujenga Thanks a lotfor the report and the PR!
I'm aware of the issue and am likely to merge your PR relatively soon 👍 

My only concern is how we could test it and verify it properly follows the upstream implementation, as we may end up adding support also for `--set-file` 🤔So please let me think for a while.
In case you have good idea(s) for that, I'd appreciate your help.

kujenga:
I see, makes sense! I have tested the myself for the issue that motivated this issue and the PR, and I did my best to mirror the implementation in Helm almost exactly, so I think the risk of having divergent behavior is fairly low, but I certainly see the point.

I wasn't aware of `--set-file`, but that definitely seems worthy of inclusion as well! Just thinking out loud, it might be nice if Helm proper exposed the configuration for these flags through an API that plugins could make use if. If that were the case, your PR here would have done this automatically: https://github.com/databus23/helm-diff/pull/82

PR #84, opened 2018-08-21
add --set-string flag to ensure proper parsing of strings

kujenga:
This resolves an issue where numeric values intended to be
used as strings could be interpreted as float64 values,
causing varius things to fail in unexpected ways.

This fix mirrors the changes in the following PR to Helm
itself: https://github.com/helm/helm/pull/3599

Closes https://github.com/databus23/helm-diff/issues/83


kujenga:
Awesome, happy to lend a hand!

PR #106, opened 2018-10-28
Feat: Added an ability to pass values from stdin

### LINKED ISSUE #101: values from stdin
I store my values in some place in base64.
It would be great if I was able to do something like this:

```
sed "s|foo|bar|g" values-template.yaml | helm install -f - stable/foo
```

`helm` can do it - https://github.com/helm/helm/issues/2709

okgolove:
Fixes #101

### DIFF
--- a/cmd/helm.go
+++ b/cmd/helm.go
@@ -32,8 +32,10 @@ func (v *valueFiles) String() string {
 func (v *valueFiles) Valid() error {
 	errStr := ""
 	for _, valuesFile := range *v {
-		if _, err := os.Stat(valuesFile); os.IsNotExist(err) {
-			errStr += err.Error()
+		if valuesFile != "-" {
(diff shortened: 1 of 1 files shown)


cmd/helm.go
```diff
@@ -144,7 +146,14 @@ func (d *diffCmd) vals() ([]byte, error) {
...
 		currentMap := map[string]interface{}{}
-		bytes, err := ioutil.ReadFile(filePath)
+
+		var bytes []byte
+ 		var err error
+ 		if strings.TrimSpace(filePath) == "-" {
+ 			bytes, err = ioutil.ReadAll(os.Stdin)
+ 		} else {
+ 			bytes, err = ioutil.ReadFile(filePath)
+ 		}
```
mumoshu: For reference: https://github.com/helm/helm/blob/581e6cdbb8e42753556ae48ccba8cef84c40f736/cmd/helm/install.go#L373-L379

PR #109, opened 2018-11-20
Use AddFlagsTLS from helm to add tls cmd options.

cgroschupp:
To enable support for the TLS environment variables (HELM_TLS_ENABLE), I replace the old tls parse logic with the AddFlagsTLS method from Helm.


cgroschupp:
@databus23 Is the pull request ok or do you have suggestions for improvement?

cgroschupp:
Fixes #111

cgroschupp:
Added support for tls-hostname #116

databus23:
Thanks and sorry for the long delay

Issue #114, opened 2018-12-13
Diff shown for tests


cablespaghetti:
Helm's test feature allows you to put definitions of test resources in `templates/test/`. However I don't want helm diff to print out these changes because they don't actually do anything in a normal `helm upgrade` command.

To reproduce this, run something like
```
helm diff --allow-unreleased upgrade keycloak-test-release stable/keycloak
```

regner:
This causes issue in Helmsman which uses helm-diff to figure out which charts need to be upgraded and marks any chart with tests as needing an upgrade.

regner:
If I understand right this PR should solve this: https://github.com/databus23/helm-diff/pull/118

cablespaghetti:
@regner Yes it annoys me using https://github.com/roboll/helmfile

I'm not sure from that PR how the feature is supposed to be used. It might not solve my problem because I probably don't want to ignore all hooks, just ones relating to tests.

regner:
Fairly new to Helm. What other hooks might you want to be kept enabled?

cablespaghetti:
I'm not sure especially when using public charts which I didn't develop. That's the problem. :(

regner:
Yeah those hooks for post install I can see being important.

mumoshu:
@regner @cablespaghetti Hey! Thanks for the continued discussion and the fix this. I appreciate it.

Do you think we can exclude tests from the diff by default? That is, make the flag named `--include-tests` and default to `--include-tests=false`, thinking tests are only run on `helm test` and usually you don't need it included in a diff.

mumoshu:
Or make `--no-tests=true` the default - in a case that we prefer consistency with the existing flag `--no-hooks=false`

regner:
I think it's safe to exclude tests by default yeah. I don't see a case where you would want tests to be included in the diff. I would just make sure it is clearly stated in the release notes incase someone is relying on it for something.

Would you like me to make the change so that it is based on `--include-tests` and defaults to false?

mumoshu:
> make sure it is clearly stated in the release notes incase someone is relying on it for something.

Sounds good! Will do so.

> make the change so that it is based on `--include-tests` and defaults to false?

Yes! I'd appreciate it if you could do so.

regner:
Sorry for the delay. I have now pushed the requested change to #119.

cablespaghetti:
Any chance of a release with this in? I'm keen to start using it. :)

PR #119, opened 2019-01-30
First pass adding option to disable test hooks specifically

regner:
This is for #114

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -29,6 +29,7 @@ type diffCmd struct {
 	resetValues      bool
 	allowUnreleased  bool
 	noHooks          bool
+	noTests          bool
 	suppressedKinds  []string
 	outputContext    int
 }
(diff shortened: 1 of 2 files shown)


databus23:
@mumoshu looks like this broke the tests, no?

regner:
Dang! I thought I ran tests but I may not have after I made the final changes. I will take a look today.

Is there a reason tests don't get run on every commit?

regner:
@databus23 the tests all pass for me. May I ask which tests are failing for you?

I just ran `go test -v ./...` in the root of the repo.

databus23:
https://circleci.com/gh/databus23/helm-diff/22

regner:
Yeah, those tests are failing and clearly should be based on the change. I don't understand why they do not fail for me... thats so add. Fixing this now.

regner:
I have opened a PR that should fix the tests.

Issue #48, opened 2018-04-26
multiple values files


mpluhar:
Hi, thanks for this awesome helm plugin. We would love to use it, but in our setup it does not work.
We specify multiple values files with duplicate keys. helm merges them, but helm diff cannot handle it.
We get errors like that below

2018/04/26 20:41:59 Error: Found duplicate key ",  ()" in manifest
2018/04/26 20:41:59 Error: Found duplicate key ",  ()" in manifest

Any chance you could look into that?

mumoshu:
@mpluhar Hey! Thanks for the report. I took some to to read code, and am still unsure what would be the actual source of the issue.

AFAICS, the `Found duplicate key` error is reported when you have two or more k8s resources with the duplicated metadata.name. Specifically, yours seem to indicate that there's k8s resources named `, ()`. Is that really the case?

adidenko:
It happens if you have comments (i.e. some headers) in your Helm chart templates that are separated by `---` (yaml document separator) from the rest of the file.

It looks like `diff` plugin loads comments as empty "manifest" and then complaints if you have more than one of those.

Example that fails:
```
# My test SA
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: test
```

Result:
```
mynamespace, ,  () has been added:
-
+ # Source: test/templates/test.yaml
+ # My test SA
```

Example that works:
```
# My test SA
#
apiVersion: v1
kind: ServiceAccount
metadata:
  name: test
```

adidenko:
The issue still exists in `2.11.0+2`. Could someone please add `bug` label?

@mumoshu here's a simple steps how to reproduce it, just go to any deployed helm chart folder and do this (create two files with only comments inside):

```
$ helm diff upgrade test .

$ echo '#TEST1' > templates/test1.yaml
$ echo '#TEST2' > templates/test2.yaml

$ helm diff upgrade test .
2018/12/11 17:37:18 Error: Found duplicate key "test, ,  ()" in manifest
2018/12/11 17:37:18 Error: Found duplicate key "test, ,  ()" in manifest
```

Some version info:
```
$ helm diff version
2.11.0+2

$ helm version
Client: &version.Version{SemVer:"v2.12.0", GitCommit:"d325d2a9c179b33af1a024cdb5a4472b6288016a", GitTreeState:"clean"}
[...]

PR #127, opened 2019-03-11
Skip yaml blocks without any metadata

travisgroth:
Fixes the variant of #48 reported by bfin, jcalahan and adidenko.

If you have templates that wind up containing only comments, they should be skipped.  This change checks to see if the metadata is completely empty to detect effectively blank yaml content.

To do this without an extraneous comparison variable, I had to unshadow the metadata type.  This resulted in a few extra cosmetic changes in the diff.

### DIFF
--- a/manifest/parse.go
+++ b/manifest/parse.go
@@ -7,7 +7,7 @@ import (
 	"log"
 	"strings"
 
-	"gopkg.in/yaml.v2"
+	yaml "gopkg.in/yaml.v2"
 	"k8s.io/helm/pkg/proto/hapi/release"
 )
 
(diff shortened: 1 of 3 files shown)


travisgroth:
@databus23 @mumoshu would love to get this merged in if you don't have any objections.  While I think the error is harmless, it would be nice to get rid of the confusing output.  If you want any changes before accepting, let me know.

ekeih:
Hey @mumoshu, first of all I want to say thanks for this super helpful helm plugin!
Is there any chance you could tag a new release that includes this fix? It would help us a lot.
Thanks in advance
Max

PR #145, opened 2019-07-03
fix the recognition of unreleased errors

### LINKED ISSUE #121: `helm diff upgrade` doesn't respect the `--allow-unreleased` flag
I'm trying to get the diff for my released which has not been released yet.

`helm diff upgrade --allow-unreleased printing chartmuseum/printing --version ~0.1.0 --namespace shared --values printing/values.yaml` returns
```
Error: "printing-beta" has no deployed releases
"printing-beta" has no deployed releases
```
Is this the expected behavior?

majewsky:
The error message is now rewritten inside Helm from "release: 'foo' not found" into "'foo' has no deployed releases", so --allow-unreleased did not work as intended anymore with newer Tillers.

Fixes #121.

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -119,7 +119,7 @@ func (d *diffCmd) run() error {
 	releaseResponse, err := d.client.ReleaseContent(d.release)
 
 	var newInstall bool
-	if err != nil && strings.Contains(err.Error(), fmt.Sprintf("release: %q not found", d.release)) {
+	if isUnreleasedErr(err, d.release) {
 		if d.allowUnreleased {
 			fmt.Printf("********************\n\n\tRelease was not present in Helm.  Diff will show entire contents as new.\n\n********************\n")
 			newInstall = true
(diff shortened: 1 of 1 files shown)


majewsky:
For reference, this is where the error message gets rewritten: https://github.com/helm/helm/blob/66d71811e843dce325feeda0a98d528fd425f92f/pkg/storage/storage.go#L129-L131

mumoshu:
@majewsky Hey! Thanks for your help.

But I'm pretty confused. Are you saying that Helm doesn't emit error messages like `release: 'foo' not found` anymore? I was thinking that's not the case though.

I do see it in recent versions of helm+tiller:

```
$ helm tiller run -- helm get foo
Installed Helm version v2.14.2
Installed Tiller version v2.14.2
Helm and Tiller are the same version!
Starting Tiller...
Tiller namespace: kube-system
Running: helm get foo

Error: release: "foo" not found
Stopping Tiller...
Error: plugin "tiller" exited with error
```

mumoshu:
And as mentioned in #121 the error message `"'foo' has no deployed releases"` is shown when your release has been failed for first installation. All you can do after the trouble is to run `helm delete --purge` and rerun installation. Now you can run `helm diff --allow-unreleased` to show diffs before the rerun as its a clean first installation.

However, with your fix I guess `helm diff --allow-unreleased` output makes the user think that they can just run "helm install", which isn't true.

PR #148, opened 2019-08-25
Improve code quality using gofmt and go lint

aananthraj:
This PR,
- Refactors code using `gofmt` and `go lint` to improve code quality. 
- Adds scripts to `gofmt` code and to update required changes
- Adds scripts to `go lint` and `go vet` code
- Adds `lint` stage before `build` to retain code quality
- Adds Godocs Badge to README.md

**preview**:
goes from [Go Report Card](https://goreportcard.com/report/github.com/databus23/helm-diff) to [Go Report Card](https://goreportcard.com/report/github.com/aananthraj/helm-diff)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,15 @@ install: build
 	cp bin/diff $(HELM_HOME)/plugins/helm-diff/bin
 	cp plugin.yaml $(HELM_HOME)/plugins/helm-diff/
 
+.PHONY: lint
+lint:
+	scripts/update-gofmt.sh
+	scripts/verify-gofmt.sh
+	scripts/verify-golint.sh
+	scripts/verify-govet.sh
+
 .PHONY: build
-build:
+build: lint
 	mkdir -p bin/
 	go build -i -v -o bin/diff -ldflags="$(LDFLAGS)"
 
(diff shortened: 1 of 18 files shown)


aananthraj:
@databus23, @mumoshu. Can this be reviewed and merged?

aananthraj:
self-reviewed and pushed the changes.

aananthraj:
> so that the method name conforms to the convention of being "verb (+ noun)".

I do agree with your point. I had to change it, as it was reported by `gofmt` that 
package name and func name are same.

mumoshu:
@aananthraj Ah good point. Then `Just `diff.Manifests` and `diff.Releases` without `Diff` makes sense to me. Would you mind changing code in that way?

It is a breaking change for anyone using helm-diff as a go library. But it isn't an intended use-case at the moment and I'll keep it until I hear from anyone requesting to bring back the go-level API compatibility. That said, please forgive me if I reverted part of this or altered the lint script to ignore errors on the two method names :)

aananthraj:
> Would you mind changing code in that way?

Sure will make those changes and update here
> It is a breaking change for anyone using helm-diff as a go library. 

https://godoc.org/github.com/databus23/helm-diff
Godocs says helm-diff is presently not being used as a golang package.
So i guess change this will have very minimalistic impact.

aananthraj:
@mumoshu made those requested changes

aananthraj:
Thanks @mumoshu for the prompt responce and support 👍

Issue #147, opened 2019-07-27
support helm v3


nwsparks:
With work on Helm 3 progressing at a steady pace, are there any plans to make this plugin work with it? Currently it seems to not be functional and exits with this:

```
Error: context deadline exceeded
context deadline exceeded
Error: plugin "diff" exited with error
helm.go:76: [debug] plugin "diff" exited with error
main.loadPlugins.func2
        /go/src/helm.sh/helm/cmd/helm/load_plugins.go:95
helm.sh/helm/vendor/github.com/spf13/cobra.(*Command).execute
        /go/src/helm.sh/helm/vendor/github.com/spf13/cobra/command.go:762
helm.sh/helm/vendor/github.com/spf13/cobra.(*Command).ExecuteC
        /go/src/helm.sh/helm/vendor/github.com/spf13/cobra/command.go:850
helm.sh/helm/vendor/github.com/spf13/cobra.(*Command).Execute
        /go/src/helm.sh/helm/vendor/github.com/spf13/cobra/command.go:800
main.main
        /go/src/helm.sh/helm/cmd/helm/helm.go:75
runtime.main
        /usr/local/go/src/runtime/proc.go:200
runtime.goexit
        /usr/local/go/src/runtime/asm_amd64.s:1337
```

mumoshu:
I'm willing to support it, but definitely appreciate it if anyone could take it and submit a PR :)

I'm sharing what I know for anyone interested.

Helm 3 seems to have a different go module name which allows us to import both helm v2 and v3 as go libraries and make helm-diff works against both versions of helm.

The part of the problem I thought difficult was how helm-diff should determine which version of helm the plugin was called by.

Fortunately, Helm v3 seems to pass a different set of envvars to plugins:

v3: https://github.com/helm/helm/blob/97e7461e41455e58d89b4d7d192fed5352001d44/pkg/plugin/plugin.go#L217
v2: https://github.com/helm/helm/blob/0e7f3b6637f7af8fcfddb3d2941fcc7cbebb0085/pkg/plugin/plugin.go#L172

So basically helm-diff should be able to determine the version of helm by checking the existence of `TILLER_HOST`.

[...]

PR #150, opened 2019-09-07
Support of helm 3

jonasrutishauser:
Full support of helm v3.0.0-beta.3 and helm 2.

I implemented this independent of #149 but it will be solution for #147 which supports all commands.

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -1,12 +1,17 @@
 package cmd
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"helm.sh/helm/pkg/action"
+	"helm.sh/helm/pkg/chart"
+	"helm.sh/helm/pkg/chart/loader"
+	"helm.sh/helm/pkg/cli/values"
+	"helm.sh/helm/pkg/getter"
 	"k8s.io/helm/pkg/helm"
 
 	"github.com/databus23/helm-diff/diff"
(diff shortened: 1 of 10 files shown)


cmd/upgrade.go
```diff
@@ -116,10 +119,16 @@ func (d *diffCmd) run() error {
 		return err
 	}
 
-	releaseResponse, err := d.client.ReleaseContent(d.release)
+	releaseResponse, err := d.deployedRelease(d.release)
+	var notDeployed string
+	if isHelm3() {
+		notDeployed = fmt.Sprintf("%q has no deployed releases", d.release)
```
mumoshu: Ah good catch! I missed this in my PR #149

mumoshu:
@jonasrutishauser Thanks for your contribution! This looks so nice 💯 

I'm now trying to build and test this myself, but had no luck so far.

`glide install && HELM_HOME=$HOME/Library/helm make install` emits a bunch of compilation errors perhaps due to inconsistent dependencies https://gist.github.com/mumoshu/1dfda422fce533493128fa93f07ef9ac

Do you have any clue on this?

jonasrutishauser:
I had many difficulties with glide too and after a `glide up` the build did not work, but after a `make bootstrap` which calls `glide install --strip-vendor` the build worked.
I even used more time with glide than with the implementation.

databus23:
@jonasrutishauser Thanks for the PR. I think we need to make a general decision how this plugin should handle the rendering part for helm3 as there are some potential compatibility issues across versions of helm that I'm worried about. I'm glad to here your thought on this comment: https://github.com/databus23/helm-diff/pull/149#issuecomment-529403225

databus23:
Closed in favour of #151

PR #153, opened 2019-10-04
Move to go modules

chenrui333:


chenrui333:
checkin

liuming-dev:
ping

mumoshu:
This seems to have been the basis of #194 and superseded by it. Thanks for your contribution!
// We had to update Makefile to keep helm version tracking supported and it isn't even covered by #194. But we can defer it to another PR

PR #160, opened 2019-10-25
fix issues in --reset-values and --reuse-values on helm v3

jonasrutishauser:
Small correction of #159


mumoshu:
@databus23 Could you cut next rc release of helm-diff with this one when you have some time?

databus23:
@mumoshu done

PR #172, opened 2019-12-12
Handle multiline strings a bit better

brimstone:
I've found diffing a file in configmap kinda terrible. This change converts strings with a "\n" sequence into a multi-line string so it's easier for difflib to do its thing.


brimstone:
@databus23 anything I need to do to get this merged?

brimstone:
@mumoshu anything I can do to help get this merged?

PR #175, opened 2019-12-27
app option to pass apiVersions to helm

a-hat:
This PR adds a new flag `--api-versions` which values are passed to `helm template`.

The flag can be used to explicitly specify the API Versions which are expected to be supported by the cluster, in order to avoid getting unexpected results with helm charts checking for `Capabilities.APIVersions`.

This feature in `helm-diff` will enable the same option flag in `helmfile` (see https://github.com/roboll/helmfile/issues/1014 and https://github.com/roboll/helmfile/pull/1046)

### DIFF
--- a/cmd/helm3.go
+++ b/cmd/helm3.go
@@ -89,7 +89,9 @@ func (d *diffCmd) template() ([]byte, error) {
 	for _, fileValue := range d.fileValues {
 		flags = append(flags, "--set-file", fileValue)
 	}
-
+	for _, apiVersion := range d.apiVersions {
+		flags = append(flags, "--api-versions", apiVersion)
+	}
 	//This is a workaround until https://github.com/helm/helm/pull/6729 is released
 	for _, apiVersion := range strings.Split(os.Getenv("HELM_TEMPLATE_API_VERSIONS"), ",") {
 		if apiVersion != "" {
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -32,6 +32,7 @@ type diffCmd struct {
 	includeTests     bool
 	suppressedKinds  []string
 	outputContext    int
+	apiVersions      []string
 }
 
 const globalUsage = `Show a diff explaining what a helm upgrade would change.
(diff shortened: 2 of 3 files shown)


a-hat:
Hey @databus23! Would you mind having a look at this PR? The corresponding feature in https://github.com/roboll/helmfile is dependent on this, it would be highly appreciated if this can be merged. Thanks!

databus23:
Hi @a-hat, thanks for the PR. I'm aware that `.Capabilities` are not working as expected in helm-diff at the moment. this is due to limitations with the `helm template` command that the diff plugin used under the hood.
I addressed this in an upstream PR https://github.com/helm/helm/pull/7054 that will be in in 3.1.0

With this PR helm diff can use `helm template --validate --is-upgrade` to correctly render the chart with `Capabilities` set according to what is actually found in the cluster.

I'm reluctant to merge this PR as it only solves the problem partly.
(e.g. `.Release.IsUpgrade` and `.Capabilities.KubeVersion` can't be set manually in the `helm template` command)
In addition it also requires to manually specify what versions are available which is error-prone.

So I kindly ask you to be patient until helm 3.1.0 is released and we can update this plugin to solve the template rendering properly.

travisghansen:
My use case will likely require the ability to manually specify versions. Not sure what the upstream fix is doing but a completely disconnected process is relevant for me.

databus23:
> but a completely disconnected process is relevant for me.

This makes no sense for this plugin. The whole point of `helm diff` is to create a diff between the currently installed release and a new (to be installed) release.
Basically create a diff between `helm get manifest` and `helm template`. This always requires a connection to the cluster to fetch the currently installed release manifest.
A disconnected process is not feasible with this plugin and never was.

travisghansen:
Good point. Lost the forest through the trees :) thinking about too many things at once I guess.

travisghansen:
@a-hat does helmfile execute diff behind the scenes during template or otherwise?

a-hat:
@databus23 Thanks for your feedback. So if I understand correctly, with helm 3.1.0, `helm diff` will work correctly with `Capabilities.APIVersions`. This is fine for me, I will close this PR.

a-hat:
> @a-hat does helmfile execute diff behind the scenes during template or otherwise?

@travisghansen `helmfile apply` first executes `helm diff` to determine if changes need to be applied. `helmfile template` executes `helm template`.

travisghansen:
Yeah ok. I misread the other thread so apologies for the noise :(

Issue #184, opened 2020-02-20
Doesn't build on fresh install


Xartos:
If I'm following the instructions for building the project there's a lot of dependencies that is missing.
After looking in the Makefile I realized that there's a bootstrap target that one can run to fetch most of the dependencies. This should be documented.
But even after those were installed there were still missing dependencies:
```
$ make install
scripts/update-gofmt.sh
scripts/verify-gofmt.sh
scripts/verify-golint.sh
scripts/verify-govet.sh
+ go vet . ./cmd/... ./manifest/...
vendor/k8s.io/api/core/v1/generated.pb.go:225:8: cannot find package "github.com/gogo/protobuf/proto" in any of:
	/home/freddeliv/go/src/github.com/databus23/helm-diff/vendor/github.com/gogo/protobuf/proto (vendor tree)
	/usr/lib/go-1.12/src/github.com/gogo/protobuf/proto (from $GOROOT)
	/home/freddeliv/go/src/github.com/gogo/protobuf/proto (from $GOPATH)
vendor/k8s.io/api/core/v1/generated.pb.go:235:8: cannot find package "github.com/gogo/protobuf/sortkeys" in any of:
	/home/freddeliv/go/src/github.com/databus23/helm-diff/vendor/github.com/gogo/protobuf/sortkeys (vendor tree)
	/usr/lib/go-1.12/src/github.com/gogo/protobuf/sortkeys (from $GOROOT)
	/home/freddeliv/go/src/github.com/gogo/protobuf/sortkeys (from $GOPATH)
vendor/k8s.io/apimachinery/pkg/labels/selector.go:26:2: cannot find package "github.com/golang/glog" in any of:
	/home/freddeliv/go/src/github.com/databus23/helm-diff/vendor/github.com/golang/glog (vendor tree)
	/usr/lib/go-1.12/src/github.com/golang/glog (from $GOROOT)
	/home/freddeliv/go/src/github.com/golang/glog (from $GOPATH)
vendor/k8s.io/apimachinery/pkg/apis/meta/v1/micro_time.go:23:2: cannot find package "github.com/google/gofuzz" in any of:
	/home/freddeliv/go/src/github.com/databus23/helm-diff/vendor/github.com/google/gofuzz (vendor tree)
	/usr/lib/go-1.12/src/github.com/google/gofuzz (from $GOROOT)
	/home/freddeliv/go/src/github.com/google/gofuzz (from $GOPATH)
[...]

PR #185, opened 2020-02-20
Added detailed exitcode flag to revision, release and rollback

Xartos:
The rollback, release and revision command has the code for using detailed exit code but the flags are not there. So I added the flags.


Xartos:
Fails because of https://github.com/databus23/helm-diff/issues/184

databus23:
Thanks.

PR #190, opened 2020-02-27
Update installation instructions

deiga:
Install should work now without `--version master`


deiga:
@mumoshu Could you review this? :)

PR #196, opened 2020-03-21
Feat: Expose post-renderer flag in the upgrade subcommand

### LINKED ISSUE #195: [feat] Allow passing post-renderer flag in the upgrade subcommand
## Feature request summary

Expose `--post-renderer` flag in the `upgrade` subcommand.

## Motivation

I've noticed that in the event of having installed a chart using the post-renderer flag, the diff is not accurate.
This happens because we are comparing a manifest rendered by helm only, but that it didn't go through the post-render process, contrarily to the released manifest.

If we do `helm upgrade "mychart" . --post-renderer ./mypostrenderer.sh`, the released manifest pipeline is:
helm rendered chart `->` post-renderer manipulation `->` manifest_1 (deployed)
After upgrading the chart, If we do: `helm upgrade "mychart" . --debug --dry-run`, the pipeline is:
helm rendered chart `->` manifest_2

To avoid this issue, the plugin would need to expose the `--post-renderer` flag and forward it to helm.

ruizink:
Fixes #195

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ Examples:
 Flags:
   -h, --help                   help for upgrade
       --detailed-exitcode      return a non-zero exit code when there are changes
+      --post-renderer string   the path to an executable to be used for post rendering. If it exists in $PATH, the binary will be used, otherwise it will try to look for the executable at the given path
       --reset-values           reset the values to the ones built into the chart and merge in any new values
       --reuse-values           reuse the last release's values and merge in any new values
       --set stringArray        set values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)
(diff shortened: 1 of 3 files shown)


acaire:
Works perfectly for me with helm `v3.1.2`:
```
helm diff upgrade --allow-unreleased --post-renderer /tmp/kustomize redis stable/redis
```
and `/tmp/kustomize` as :
```
#!/bin/bash
cat <&0 > all.yaml
kustomize build . && rm all.yaml
```

and various yaml:
```
# kustomization.yaml
resources:
- all.yaml
patches:
- replicas.yaml
```
```
# replicas.yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: redis-slave
spec:
  replicas: 200
```

AntonioMeireles:
@databus23 

Hi, 

what's holding this specific feature from landing into a formal release ? is there anything the community could help in order to speed up things ;-) ?

thanks in advance!

AntonioMeireles:
Thanks all for getting this in ! what a about a new version now to match stock helm's ;-) 

All the best!

PR #206, opened 2020-04-16
Change to use sh instead of bash

### LINKED ISSUE #204: can't execute 'bash'
My case was installing helm-diff to an alpine container without bash.
I go the error from the subject.

As far as I understand this is because of this shebang https://github.com/databus23/helm-diff/blob/master/install-binary.sh#L1

Should we use `/usr/bin/env sh` instead of bash?

deiga:
Ensure POSIX compatibility

Fixes #204

### DIFF
--- a/install-binary.sh
+++ b/install-binary.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/sh sh
+#!/usr/bin/env sh
 
 # Shamelessly copied from https://github.com/technosophos/helm-template
 
@@ -8,7 +8,7 @@ export GREP_COLOR="never"
 
 HELM_MAJOR_VERSION=$(helm version --client --short | awk -F '.' '{print $1}')
 
-if [ "$HELM_MAJOR_VERSION" = "v2" ]; then
+if [ "$HELM_MAJOR_VERSION" = "Client: v2" ]; then
   : ${HELM_PLUGIN_DIR:="$(helm home --debug=false)/plugins/helm-diff"}
 fi
 
--- a/install-binary.sh
+++ b/install-binary.sh
@@ -6,6 +6,12 @@ PROJECT_NAME="helm-diff"
 PROJECT_GH="databus23/$PROJECT_NAME"
 export GREP_COLOR="never"
 
+HELM_MAJOR_VERSION=$(helm version --client --short | awk -F '.' '{print $1}')
+
+if [ "$HELM_MAJOR_VERSION" = "v2" ]; then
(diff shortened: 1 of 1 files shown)


install-binary.sh
```diff
@@ -1,23 +1,21 @@
...
-: ${HELM_PLUGIN_DIR:="$(helm home --debug=false)/plugins/helm-diff"}
+export GREP_COLOR="never"
 
 # Convert the HELM_PLUGIN_DIR to unix if cygpath is
 # available. This is the case when using MSYS2 or Cygwin
 # on Windows where helm returns a Windows path but we
 # need a Unix path
 
-if type cygpath > /dev/null 2>&1; then
+if type cygpath >/dev/null 2>&1; then
```
mumoshu, line 14: Is this a cosmetic change, or actually required for POSIX compliance?

install-binary.sh
```diff
@@ -1,23 +1,21 @@
-#!/usr/bin/env bash
+#!/usr/bin/sh sh
 
 # Shamelessly copied from https://github.com/technosophos/helm-template
 
 PROJECT_NAME="helm-diff"
 PROJECT_GH="databus23/$PROJECT_NAME"
-GREP_COLOR="never"
-
-: ${HELM_PLUGIN_DIR:="$(helm home --debug=false)/plugins/helm-diff"}
```
mumoshu, line 9: We do use HELM_PLUGIN_DIR in `installFile`. Should we just revert this or just make it POSIX-compliant?

install-binary.sh
deiga: Cosmetic change

install-binary.sh
deiga: @mumoshu AFAIK the `:` denotes the line as pretty much a comment and thus would not run. As `helm home` is not supported in helm3

mumoshu: @deiga Thanks. I thought so at first, but my tiny experiment revealed that this seems to have actual side-effect of assigning the value.

```
$ bash -c ': ${FOO:=somevalue}; echo $FOO'
somevalue
```

> As helm home is not supported in helm3

Good catch! Then it can be said that this can safely be removed for helm 3. Not yet sure that's alright for helm 2 though.

deiga: Right. Should I add a conditional for helm2?

mumoshu: @deiga That would be awesome! It will allow us to move forward without deciding if it's necessary or not right now.

deiga:
@mumoshu Sorry, this took me forever to get back to. It should now work for both Helm v2 and v3

okgolove:
Any chances this one will be merged?

Issue #209, opened 2020-04-29
Support helm3 --disable-openapi-validation


ZeeFiix:
Is there support for, or can this be added for this helm3 flag? --disable-openapi-validation

The same question was raised for helmfile.
https://github.com/roboll/helmfile/issues/1167

helm3 has introduced this switch to be used as workaround as long openshift has these problems.
It would really be helpfull it it also can be used with helm diff plugin

PR #210, opened 2020-05-03
Add support for Helm 3 --disable-openapi-validation flag

Kirill-Garbar:
Added disable-openapi-validation helm3 flag support to disable rendered templates validation with the Kubernetes OpenAPI Schema.
https://github.com/helm/helm/pull/7570
It fixes https://github.com/databus23/helm-diff/issues/209

### DIFF
--- a/cmd/helm3.go
+++ b/cmd/helm3.go
@@ -81,7 +81,7 @@ func getChart(release, namespace string) (string, error) {
 	return string(out), nil
 }
 
-func (d *diffCmd) template(isUpgrade bool, disableOpenAPIValidation bool) ([]byte, error) {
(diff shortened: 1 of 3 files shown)


cmd/helm3.go
```diff
@@ -129,6 +129,10 @@ func (d *diffCmd) template(isUpgrade bool) ([]byte, error) {
 		flags = append(flags, "--is-upgrade")
 	}
 
+	if disableOpenAPIValidation {
```
mumoshu, line 132: You should just use `if d.disableOpenAPIValidation {` instead of passing the extra arg and referencing it

Kirill-Garbar: @mumoshu Thank you for your review. Fixed and resolved conflicts.

PR #221, opened 2020-06-30
Add support for custom output formats

semoac:
Hi!

I was using these changes to format the output of helm-diff and I think it may be useful to someone else.
 
I use these formats to help me make decisions during my CI/CD pipeline on whatever or not the changes should be applied unsupervised. They also help me reduce the visual noise when `helmfile`  has to process a lot of releases.

I also been using the "template" output format to feed data to external system that needs to know what resource could be affected on upgrades.

This PR do not fix anything. I'm just sharing code. Please feel free to reject it without explanation.

### Simple with helmfile
Screen Shot 2020-06-29 at 15 54 18

### Template with a json like output
Screen Shot 2020-06-29 at 02 31 13


mumoshu:
@semoac Also - would you mind if we reused your screencaptures in our README?

semoac:
> @semoac Also - would you mind if we reused your screencaptures in our README?

Sure. Go ahead!

PR #226, opened 2020-07-20
feat: JSON output

mumoshu:
This adds the JSON output on top of #221, which can be enabled by passing `--output json`. Internally it works exactly the same as the `template` output format with the default template.

```
********************

        Release was not present in Helm.  Diff will show entire contents as new.

********************
[{
  "api": "v1",
  "kind": "Secret",
  "namespace": "default",
  "name": "mysql-1567775891",
  "change": "ADD"
},{
  "api": "v1",
  "kind": "ConfigMap",
  "namespace": "default",
  "name": "mysql-1567775891-test",
  "change": "ADD"
},{
  "api": "v1",
  "kind": "PersistentVolumeClaim",
  "namespace": "default",
  "name": "mysql-1567775891",
  "change": "ADD"
},{
  "api": "v1",
  "kind": "Service",
  "namespace": "default",
  "name": "mysql-1567775891",
  "change": "ADD"
},{
  "api": "apps",
  "kind": "Deployment",
  "namespace": "default",
  "name": "mysql-1567775891",
  "change": "ADD"
}]
```

Also note that I've modified the default template a bit so that it looks more like helm's output. See `helm list -o json` for example.

Ref https://github.com/databus23/helm-diff/pull/221#pullrequestreview-451199381

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,13 +5,13 @@ PKG:= github.com/databus23/helm-diff
 LDFLAGS := -X $(PKG)/cmd.Version=$(VERSION)
 
 # Clear the "unreleased" string in BuildMetadata
-LDFLAGS += -X $(PKG)/vendor/k8s.io/helm/pkg/version.BuildMetadata=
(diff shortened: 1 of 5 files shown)


mumoshu:
@semoac Would you be okay with the update on the default template?

semoac:
> @semoac Would you be okay with the update on the default template?

Yes! Thank you!

PR #233, opened 2020-08-10
enable semantic import versioning for go mod compat

tariq1890:
I forgot to add the semantic version in the gomod in this PR. 

As per go mod rules, if we are using a version greater that v1, we need suffix the go module name with that version. 

This PR fixes that


tariq1890:
@mumoshu FYI

tariq1890:
@databus23

Issue #144, opened 2019-06-11
Option to ignore line endings


michaelgeorgeattard:
When a particular upgrade/release has been made with CRLF line ending template files, and a diff is made with a LF line ending version of the same template files, `helm diff` shows the line ending change as a difference (as expected).

Is it currently possible to ignore line ending changes when generating the diff?

ranjanashish:
It would be great if helm-diff has certain flags which users can use to ignore differences in line endings and, also, comments (since comments are not going to affect the workload).

mebays:
I just started using helm diff and this would be a great feature to incorporate.

shimodatkh:
I have just faced this problem. It would be great if an option to ignore line endings was implemented!

mumoshu:
Could anyone submit a pull request for this? This sounds like a good addition to helm-diff.

mumoshu:
This has been added via #237 thanks to @bakito 😄

PR #237, opened 2020-08-24
add support for strip-trailing-cr

bakito:
This PR add support to strip the trailing cr from charts that have been deployed wit different line endings.

See #144

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -39,6 +39,7 @@ type diffCmd struct {
 	postRenderer             string
 	output                   string
 	install                  bool
+	stripTrailingCR          bool
 }
 
 func (d *diffCmd) isAllowUnreleased() bool {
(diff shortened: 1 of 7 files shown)


amorozov87:
Can somebody merge this request?

bakito:
Can someone please have a look at this?

cmd/upgrade.go
```diff
@@ -270,7 +272,8 @@ func (d *diffCmd) run() error {
 		}
 	}
 
-	seenAnyChanges := diff.Manifests(currentSpecs, newSpecs, d.suppressedKinds, d.showSecrets, d.outputContext, d.output, os.Stdout)
+	println("#######################################################################################################################################*******************************************************")
```
mumoshu, line 275: Can we remove this as an irrelevant change?

mumoshu (approved):
Sorry for the delayed response.

Just one comment on a seemingly redundant println. It would be great if you could confirm that it does look like an unnecessary change.

I'll merge this and revert the println addition only after a few days anyway.
Thanks for your contribution @bakito!

bakito:
Hi @mumoshu 
I remove the debug line, sorry I missed that.

mumoshu:
@bakito Thanks again for your support! Merging.

PR #240, opened 2020-09-10
Automatically color output based on whether output is a terminal

misberner:
In the absence of an explicit `--no-color` flag specification, try to infer whether the standard output is a terminal, and only emit colored output when it is.
This only affects invocations without the flag, all of `--no-color`, `--no-color=false`, and `--no-color=true` retain their original semantics.

See example:
Screen Shot 2020-09-10 at 23 54 47

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,10 +1,11 @@
 package cmd
 
 import (
+	"os"
+
 	"github.com/mgutz/ansi"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


misberner:
@mumoshu any interest in this?

misberner:
@mumoshu ping on this. Is this project still maintained?

mumoshu (approved):
LGTM. Thanks for your patience and contribution @misberner ☺️

PR #243, opened 2020-10-07
Ignore unknown flags for upgrade

johanfleury:
Hello.

This PR is meant for people like me who want to use `--atomic` and also like to run `helm upgrade` by simply removing `diff` from the command line.

It allows to run `helm diff upgrade --atomic ...` without getting `Error: unknown flag: --atomic`.


johanfleury:
I love that kind of bot…

mumoshu:
Thanks for this! Let me link this to #278 and discuss there.

PR #261, opened 2021-02-16
diff/report: fix key regexp

eonpatapon:
The current regexp doesn't match all possible apis, for example
`cert-manager.io`.

As there is a closing `)` we can simply take everything before it.


mumoshu:
@eonpatapon Hey! Thanks for your contribution. Just read your explanation and the code change and they look good to me on their own. but I'm still not sure what was the real impact of the issue.

Did we ever have any error related to this?

eonpatapon:
When running helm diff it will output errors like: `error processing report entry`

eonpatapon:
Sad...

PR #268, opened 2021-03-12
Allow passing the chart repo to helm template

aackerman:


mumoshu (approved):
Wow, I've never realized helm had this option and I'm entirely unsure when this is supposed to be used, but I learned that it works like:

```
helm template --repo https://charts.helm.sh/stable envoy
```

Anyway, we'd better have this to make it consistent with `helm template`.

LGTM. Thanks a lot for your contribution @aackerman!

aackerman:
@mumoshu awesome, thanks for getting this merged. This is important for remote repos that aren't already local.

mumoshu:
@aackerman Thanks for the info! So, this just frees you from being required to run `helm repo add` beforehand, right?

aackerman:
@mumoshu yep that's definitely what I'm using it for

mumoshu:
@aackerman Got it. Thanks for clarifying ☺️

PR #270, opened 2021-03-20
fix: correct helm binary in install script for environment that have both helm 2 and 3 installed

astorath:
This fixes `install-binary.sh` to use correct helm binary

### DIFF
--- a/install-binary.sh
+++ b/install-binary.sh
@@ -6,9 +6,9 @@ PROJECT_NAME="helm-diff"
 PROJECT_GH="databus23/$PROJECT_NAME"
 export GREP_COLOR="never"
 
-HELM_MAJOR_VERSION=$(helm version --client --short | awk -F '.' '{print $1}')
+HELM_MAJOR_VERSION=$("${HELM_BIN}" version --client --short | awk -F '.' '{print $1}')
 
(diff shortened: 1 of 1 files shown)


deiga:
@astorath Can you describe what the problem is right now? Why is this fix needed?

astorath:
@deiga Sure, the problems are, when you have helm binary named differently (helm2 or helm3 for instance):

1. Installation does not work sometimes (`HELM_PLUGIN_DIR` resolves to `/plugins/helm-diff`) because root fs might be readonly
2. `/plugins/helm-diff` is created polluting filesystem
3. The latest version is installed always (because `git describe` fails on empty `/plugins/helm-diff`)

PS: I also have a question about that. Would you be so kind to release special `v2.11.0+6` tagged v2 release containing the same fix. I have github fork for that, but it would be nice to have the same issue fixed for helm2 in original repo.

mumoshu (approved):
Good catch! I've tried my best to rephrase the PR title to better capture the purpose of this fix. Thanks for your contribution @astorath ☺️

mumoshu:
@deiga Thanks for your help on communication! I wouldn't have understood this without your conversation with @astorath

PR #276, opened 2021-04-19
Added Source Installation Documentation

markscamilleri:
## Changes:
 - Added documentation to install from source (useful if no binaries are compatible)
 - Added a dependency on the `build` target in the Makefile's `install/helm3` target to allow one-line installation

Ref #275


mumoshu (approved):
@markscamilleri Hey! This might be super useful for future contributors and advanced users who want to customize helm-diff. Thanks a lot for your contribution @markscamilleri ☺️

PR #282, opened 2021-05-28
Ensure an initial separator is present in the yaml manifest

tommasopozzetti:
The current parsing logic assumes that the manifest containing all resources will start with a yaml separator and therefore it
discards the first token when parsing. While this is the default helm behavior, custom postprocessing of the manifests could lead to the initial separator missing. Helm will still accept this, since the resulting manifest is still a valid yaml file, but the diff plugin will always discard the first resource in such case, showing it as a missing resource in the diff.

This commit adds an extra check to ensure the assumption of initial yaml separator is always valid, by adding it in case it is missing.


dirathea:
I've just encountered the same issue. The first manifest is missing and the diff returns an incorrect result.

I've tested the changes and it works perfectly. @databus23 can you check this ?

PR #287, opened 2021-06-07
Use cygpath for HELM_BIN

jkroepke:
Since https://github.com/databus23/helm-diff/commit/7f10f255cdea7270b7916e103b532fbb1687ea69 HELM_BIN instead helm is used.

If helm plugin install is running on cygwin, an error occurred:

```
# Installed plugin: secrets
...
# D:\a\helm-secrets\helm-secrets\tests\.tmp\cache\CYGWIN_NT-10.0\helm\home\helm\plugins\helm-diff/install-binary.sh: line 9: C:\hostedtoolcache\windows\helm\3.6.0\x64\windows-amd64\helm.exe: command not found
...

# Error: plugin install hook for "diff" exited with error
```

Like HELM_PLUGIN_DIR, HELM_BIN needs to be translated on cygwin, too.


jkroepke:
.

mumoshu (approved):
Thanks for your contribution!

Honestly speaking I don't use cygwin and have no expertise on it but from what I can see from code I tried to rebase changes onto the lastest master. I hope it works.

mumoshu:
It would be great if anyone could enahnce our CI workflow to cover cygwin (if it's possible).

PR #289, opened 2021-06-15
use helm upgrade to simulate an upgrade

hypnopotamus:
use the upgrade command to simulate an upgrade unless cluster contact is disabled by another option, then use template

refs roboll/helmfile#1879
refs helm/helm#9797

- in helm/helm#8137 the discussion seemed to lead to "`helm template` should not render with cluster contact for security reasons even with `--validate`, `helm install` and `helm upgrade` are assumed to have cluster contact and should render with it, even on a `--dry-run`"
- using template compared to current state to simulate an upgrade and produce a diff has a different effect than using [the relevant parts of] an upgrade --dry-run
  - specifically what I found was the behavior of the lookup template function returning nil as a default in disconnected scenarios => helm/helm#9797 would allow `lookup` to contact the cluster in `install --dry-run` and `upgrade --dry-run`

### DIFF
--- a/cmd/helm3.go
+++ b/cmd/helm3.go
@@ -81,8 +81,7 @@ func getChart(release, namespace string) (string, error) {
 	return string(out), nil
 }
 
-func (d *diffCmd) template(isUpgrade bool) ([]byte, error) {
-	flags := []string{}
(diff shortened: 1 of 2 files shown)


gorylpe:
It would be useful for me, always seeing diff for generated secrets with lookup is very annoying

PR #291, opened 2021-06-16
helm home command not available

esteganobvio:


install-binary.sh
```diff
@@ -8,7 +8,7 @@ export GREP_COLOR="never"
 
 HELM_MAJOR_VERSION=$("${HELM_BIN}" version --client --short | awk -F '.' '{print $1}')
 
-: ${HELM_PLUGIN_DIR:="$("${HELM_BIN}" home --debug=false)/plugins/helm-diff"}
+: ${HELM_PLUGIN_DIR:="$HOME/.local/share/helm/plugins/helm-diff"}
```
mumoshu, line 11: Hey! Thanks for the PR. I'm still curious in what scenario this will work or not. Anyway, on which os and circumstance is `$HOME/.local/share/helm` used as the helm home? I usually use macOS and Ubuntu but haven't seen this

PR #269, opened 2021-03-18
Use a mktemp random directory for installation

keymon:
Do not hardcode a static directory for temporary files, as it
might cause problems when running in a multi-user system
(ie: orphan files owned by other users).

Use mktemp -d instead.

Fixes #241

PR #301, opened 2021-08-16
Use a mktemp random directory for installation (repeated PR)

### LINKED ISSUE #241: Failed to create the file /tmp/helm-diff.tgz: Permission denied
This happens with both helm v2.14.3 and v3.3.0

Steps to reproduce:

    touch /tmp/helm-diff.tgz
    chmod 000 /tmp/helm-diff.tgz
    helm plugin install https://github.com/databus23/helm-diff

Expected results:

    helm downloads the appropriate tarball, unpackages it, puts the contents in their places in $HOME/.helm/
    helm removes the temporary file that was downloaded after using it

Seen instead:

Warning: Failed to create the file /tmp/helm-diff.tgz: Permission denied
curl: (23) Failed writing body (0 != 16360)
Failed to install helm-diff

The "Steps to reproduce" are a contrived scenario, but I encountered it naturally when working on a machine I share with other users. In my case, another user launched helm plugin install weeks before I did.

Workaround:
    sudo rm /tmp/helm-diff.tgz

Subsequent problem:

    tar: diff/LICENSE: Cannot open: File exists
    tar: diff/bin: Cannot utime: Operation not permitted
    tar: diff/plugin.yaml: Cannot open: File exists
    tar: diff/README.md: Cannot open: File exists
    tar: diff/bin/diff: Cannot open: File exists
    tar: diff: Cannot utime: Operation not permitted
    tar: Exiting with failure status due to previous errors
    Failed to install helm-diff

Workaround step two:

    sudo rm -rf /tmp/helm-diff/

keymon:
> Reissuing the PR https://github.com/databus23/helm-diff/pull/269 closed by stale bot
> Updated with master

Do not hardcode a static directory for temporary files, as it
might cause problems when running in a multi-user system
(ie: orphan files owned by other users).

Use mktemp -d instead.

Fixes #241
This was already tested in #269


ohartl (changes requested):
`rmTempDir` is a nice addition for cleanup, but a call to that at the end seems to be missing. Thus temp files are only cleaned up in case of a failure currently. Or is `fail_trap` naming wrong?

keymon:
It is not stale, it is just not getting attention. Needs merging :)

keymon:
> rmTempDir is a nice addition for cleanup, but a call to that at the end seems to be missing. Thus temp files are only cleaned up in case of a failure currently. Or is fail_trap naming wrong?

yep, the trap is on `EXIT`, so it will be always executed. 

Maybe `exit_trap` is a better name

ohartl:
> > rmTempDir is a nice addition for cleanup, but a call to that at the end seems to be missing. Thus temp files are only cleaned up in case of a failure currently. Or is fail_trap naming wrong?
> 
> yep, the trap is on `EXIT`, so it will be always executed.
> 
> Maybe `exit_trap` is a better name

Sorry I was confused there for a second, thanks for clearing that up.

Its now not only waiting for a failure anymore, so naming is bit wrong. I think `exit_trap` is nice! But removing the dir in in a trap is nice.

ohartl:
@databus23 Needs merging

mumoshu (approved):
LGTM. Thanks a lot for your contribution ☺️

PR #307, opened 2021-09-14
Fix for Install-Binary script

### LINKED ISSUE #244: plugin does not install if helm is helm 3 but run from helm 2
The install-binary.sh script uses the `helm home` command which only exists in helm 3. Interestingly, if either helm 2 XOR helm 3 are installed on my system, the plugin installs fine. However if both versions of helm are on my system, and `helm` is actually helm 3, and helm 2 has some other name, then the plugin no longer installs for helm 2:
`helm version` shows helm 3
`helm plugin install .../helm-diff` works
`helm2 version` shows helm 2
`helm2 plugin install .../helm-diff` fails because 'home' is not a recognized command

The error happens because `install-binary.sh` calls `helm home` which in the above scenario actually calls helm 3, which does not support `home` command (I'm guessing that helm 3's `plugin install` does not use that script otherwise it would fail too). 

Is there a reason that `$(helm home)` cannot be replaced by `$HELM_HOME`?

Hammond95:
The script failed because of some conditions assumed true, fixed as follow:

* if HELM_BIN is not set the just uses helm path
* helm home command doesn't seem to exist on helm3.x, using a default path if this is the case
* tgz downloaded file wasn't extracted properly

resolves #244

### DIFF
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -148,7 +148,7 @@ func Releases(oldIndex, newIndex map[string]*manifest.MappingResult, suppressedK
 	return Manifests(oldIndex, newIndex, suppressedKinds, showSecrets, context, output, stripTrailingCR, to)
 }
 
-func diffMappingResults(oldContent *manifest.MappingResult, newContent *manifest.MappingResult, stripTrailingCR bool ) []difflib.DiffRecord {
+func diffMappingResults(oldContent *manifest.MappingResult, newContent *manifest.MappingResult, stripTrailingCR bool) []difflib.DiffRecord {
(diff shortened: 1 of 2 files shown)


Hammond95:
@mumoshu Sorry to bother you, any chance you can give a look at my PR? Also, I would like to help maintain this project, would this be possible?

mumoshu (approved):
We lack comprehensive test cases to verify this doesn't break anything. But it looks good to me so let's try! Thanks for your patience and contribution ☺️

mumoshu:
I've verified that `helm plugin install` on helm-diff does work without `--version` flag after this change. Probably it's good.

Issue #308, opened 2021-09-21
bug: update downloads original install version


bassco:
## The issue

The installer script uses the same logic for install or update and therefore installs the original version upon update.

https://github.com/databus23/helm-diff/blob/master/install-binary.sh#L71

## How I tested:
 
```rm -rf ~/Library/helm/plugins/helm-diff
helm plugin install https://github.com/databus23/helm-diff --version v3.1.1
helm plugin list |grep -E "NAME|^diff"
NAME            	VERSION	DESCRIPTION
diff            	3.1.1  	Preview helm upgrade changes as a diff

helm plugin update diff
Downloading https://github.com/databus23/helm-diff/releases/download/v3.1.1/helm-diff-macos.tgz
.....
```
## What I was expecting:
```
helm plugin update diff
Downloading https://github.com/databus23/helm-diff/releases/download/v3.1.3/helm-diff-macos.tgz
.....
```

## Some testing:

```
helm version --client --short
v3.2.4+g0ad800e
cd ~/Library/helm/plugins/helm-diff
git describe --tags --exact-match
v3.1.1
git rev-parse HEAD
5018bd7cd7312c5a33f75dfa9748cba7b0d5dc6f
git fetch --tags
git rev-parse HEAD
5018bd7cd7312c5a33f75dfa9748cba7b0d5dc6f
git describe --tags --exact-match
v3.1.1
git describe --tags `git rev-list --tags --max-count=1`
v3.1.3
git checkout v3.1.3
helm plugin update diff
Downloading https://github.com/databus23/helm-diff/releases/download/v3.1.3/helm-diff-macos.tgz
```

## Root cause:
https://github.com/databus23/helm-diff/blob/master/plugin.yaml#L11

The installer script needs to known which mode it is running in. 

```yaml
hooks:
  install: "$HELM_PLUGIN_DIR/install-binary.sh"
  update: "$HELM_PLUGIN_DIR/install-binary.sh"
```

- install mode; use the current vcs tag
- update mode; get the latest remote tag 

This can be achieved by prefixing the `update` command with an ENV var or passing a parameter, the common installer script can differentiate which mode it is in.

bassco:
@databus23 are you able to review this change?

mumoshu:
#309 has been merged! Thanks for your contribution @bassco ☺️

PR #309, opened 2021-09-21
fix(upgrade): fix upgrade logic in install script

bassco:
Fixes the update installing the original install version as detailed in https://github.com/databus23/helm-diff/issues/308


mumoshu (approved):
LGTM. Thanks for your contribution!

PR #320, opened 2021-12-31
feat: --find-renames for discovering renames based on content

fwiesel:
If we rename objects, they will appear as
add and remove actions, which makes it difficult
to assess the semantic delta, which may only be
a simple renaming.

The current implementation creates potentially
a diff between every add/remove item, which
means O(n^2) runtime. So it should only be
enabled if such changes are rare

### DIFF
--- a/cmd/rollback.go
+++ b/cmd/rollback.go
@@ -17,14 +17,10 @@ type rollback struct {
 	release            string
 	client             helm.Interface
 	detailedExitCode   bool
-	suppressedKinds    []string
 	revisions          []string
-	outputContext      int
 	includeTests       bool
-	showSecrets        bool
-	output             string
-	stripTrailingCR    bool
 	normalizeManifests bool
+	diff.Options
 }
 
 const rollbackCmdLongUsage = `
@@ -57,9 +53,7 @@ func rollbackCmd() *cobra.Command {
 				return err
 			}
 
-			if q, _ := cmd.Flags().GetBool("suppress-secrets"); q {
-				diff.suppressedKinds = append(diff.suppressedKinds, "Secret")
-			}
+			ProcessDiffOptions(cmd.Flags(), &diff.Options)
 
 			diff.release = args[0]
 			diff.revisions = args[1:]
@@ -76,15 +70,10 @@ func rollbackCmd() *cobra.Command {
 		},
 	}
 
-	rollbackCmd.Flags().BoolP("suppress-secrets", "q", false, "suppress secrets in the output")
-	rollbackCmd.Flags().BoolVar(&diff.showSecrets, "show-secrets", false, "do not redact secret values in the output")
 	rollbackCmd.Flags().BoolVar(&diff.detailedExitCode, "detailed-exitcode", false, "return a non-zero exit code when there are changes")
-	rollbackCmd.Flags().StringArrayVar(&diff.suppressedKinds, "suppress", []string{}, "allows suppression of the values listed in the diff output")
-	rollbackCmd.Flags().IntVarP(&diff.outputContext, "context", "C", -1, "output NUM lines of context around changes")
 	rollbackCmd.Flags().BoolVar(&diff.includeTests, "include-tests", false, "enable the diffing of the helm test hooks")
-	rollbackCmd.Flags().StringVar(&diff.output, "output", "diff", "Possible values: diff, simple, template. When set to \"template\", use the env var HELM_DIFF_TPL to specify the template.")
-	rollbackCmd.Flags().BoolVar(&diff.stripTrailingCR, "strip-trailing-cr", false, "strip trailing carriage return on input")
 	rollbackCmd.Flags().BoolVar(&diff.normalizeManifests, "normalize-manifests", false, "normalize manifests before running diff to exclude style differences from the output")
+	AddDiffOptions(rollbackCmd.Flags(), &diff.Options)
 
 	rollbackCmd.SuggestionsMinimumDistance = 1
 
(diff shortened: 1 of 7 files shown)


fwiesel:
I do not see a way around the O(n^2) runtime, except with some heuristics. But I think in most use-cases that is acceptable.
What I would like to change before considering it ready for review is:
- max-content-delta: That should be more properly explained, and possibly replaced with something more intuitive.
- Missing Test: At least a simple test case for the new feature should be implemented

diff/diff_test.go
```diff
@@ -195,29 +195,61 @@ metadata:
...
+		}
+
+		require.Equal(t, `default, nginx, Deployment (apps) has changed:
+
+  apiVersion: apps/v1
+  kind: Deployment
+  metadata:
+-   name: nginx
++   name: nginx-renamed
+
```
mumoshu, line 245: Would you mind adding one more test for the case that there's change in `metadata.name` plus there're one update, one deletion, and one insertion under `spec`?
I think that would make this feature extremely hard to get regression in the future!

cmd/rollback.go
```diff
@@ -81,6 +82,7 @@ func rollbackCmd() *cobra.Command {
 	rollbackCmd.Flags().BoolVar(&diff.detailedExitCode, "detailed-exitcode", false, "return a non-zero exit code when there are changes")
 	rollbackCmd.Flags().StringArrayVar(&diff.suppressedKinds, "suppress", []string{}, "allows suppression of the values listed in the diff output")
 	rollbackCmd.Flags().IntVarP(&diff.outputContext, "context", "C", -1, "output NUM lines of context around changes")
+	rollbackCmd.Flags().Float32VarP(&diff.maxChangeFraction, "max-content-change-fraction", "D", 0, "maximum fraction of changed content as lines added + removed compared to total lines in a diff for considering it a moved resource")
```
mumoshu, line 85:
```suggestion
	rollbackCmd.Flags().Float32VarP(&diff.findRenames, "find-renames", "D", 0, "Enable rename detection if set to any value greater than 0. If specified, the value denotes the maximum fraction of changed content as lines added + removed compared to total lines in a diff for considering it a rename")
```

How about using `find-renames` like the similar feature exists in `git diff`?

```
       -M[<n>], --find-renames[=<n>]
           Detect renames. If n is specified, it is a threshold on the similarity index (i.e. amount of addition/deletions compared to the file's size). For example, -M90% means Git should consider a delete/add pair
           to be a rename if more than 90% of the file hasn't changed. Without a % sign, the number is to be read as a fraction, with a decimal point before it. I.e., -M5 becomes 0.5, and is thus the same as -M50%.
           Similarly, -M05 is the same as -M5%. To limit detection to exact renames, use -M100%. The default similarity index is 50%.
```

cmd/upgrade.go
```diff
@@ -276,7 +278,7 @@ func (d *diffCmd) run() error {
 		}
 	}
 
-	seenAnyChanges := diff.Manifests(currentSpecs, newSpecs, d.suppressedKinds, d.showSecrets, d.outputContext, d.output, d.stripTrailingCR, os.Stdout)
+	seenAnyChanges := diff.Manifests(currentSpecs, newSpecs, d.suppressedKinds, d.showSecrets, d.outputContext, d.maxChangeFraction, d.output, d.stripTrailingCR, os.Stdout)
```
mumoshu, line 281: Can you make the rename detection get triggered only when `newInstall` is `false`? In L226-L228 we already call helm to detect if it's a diff for an upgrade(=there can be renames) or not(=there's no rename by the nature of it's a new release).

That way, I think O(N^2) issue won't affect new install badly.

cmd/rollback.go
fwiesel: It is indeed inspired by `find-renames`... 

 I was doubting if the expectation of a `find-renames` would be to only discover changes of `metadata.name` and possibly `metadata.namespace`, but not the kind. 
My problem though is, that the proposal also detects changes of the `Kind` (e.g. `Deployment` -> `StatefulSet`).

But since your intuition is that it is the same feature, I may be overthinking things here.

In that light, what is your opinion? If you still say it is more intuitive to use the same parameters, I will change it. I wasn't sure myself.

cmd/upgrade.go
fwiesel: Sure, I'll change it

diff/diff_test.go
fwiesel: Sure, I'll add them.

cmd/rollback.go
mumoshu: > My problem though is, that the proposal also detects changes of the Kind (e.g. Deployment -> StatefulSet).

Ah I missed this part. To me, it seems like it might benefit from having another flag to detect kind change, like `--find-kind-changes`? But not sure if that harmonizes with `--find-renames` then...

BTW, can we just omit the kind change detection? In most cases a kind change would result in a huge diff in spec and I'd say changing from a deployment to statefulset or vice versa doesn't happen so often. Rename detections sounds useful in more cases than kind change detection and complicating it for less used feature doesn't seem to pay off

fwiesel: Sure, I will restrict it. That also reduces also the search-space. I just thought this is more simple to implement and covers more use-cases.

fwiesel:
When implementing the tests, I was running into some minor issues I couldn't explain, so I split up the changes into several commits and refactored the code to limit the change required to implement the functionality.
I hope, it makes it better understandable what changed and why.

mumoshu (approved):
LGTM. Thanks a lot for your efforts!

PR #334, opened 2022-01-09
Start using golang 1.17.5 to build release

Sajfer:
docker-run-release was still building using golang 1.13.3. Should use 1.17 to match the rest of the project.


mumoshu (approved):
Thanks for your contribution! LGTM.

Issue #322, opened 2022-01-05
Installation fails


rkeidar:
Our CI workflow install the diff plugin. Since an hour ago it start failing (commit 4790fb13b990e7cd146211ef22cb0f5fbbba4c57)
We are running:  

helm plugin install https://github.com/databus23/helm-diff

Error:

awk: cmd. line:1: warning: regexp escape sequence `\"' is not a known regexp operator
Downloading 
curl: (3) URL using bad/illegal format or missing URL
Failed to install helm-diff
	For support, go to https://github.com/databus23/helm-diff.
Error: plugin install hook for "diff" exited with error
Error: Process completed with exit code 1.

michaeljguarino:
I'm seeing this as well

rkeidar:
I work fine when you run it with specific version:
 helm plugin install https://github.com/databus23/helm-diff --version 3.1.3

damienpontifex:
I believe it is that latest commit from #267 
Running `helm plugin install https://github.com/databus23/helm-diff` locally fails. 
Whereas `helm plugin install https://github.com/databus23/helm-diff --version 97f77494396694beb878128430205888a91d70b4` succeeds (that is the second to last commit)

damienpontifex:
I think the `grep $ARCH` from https://github.com/databus23/helm-diff/blob/4790fb13b990e7cd146211ef22cb0f5fbbba4c57/install-binary.sh#L80-L84 is returning failure exit code because the latest release (3.1.3) doesn't have architecture specific artifacts (only OS specific) and hence `DOWNLOAD_URL` is empty after `getDownloadURL` call

jBouyoud:
Seems need a new [release](https://github.com/databus23/helm-diff#release-1)

mumoshu:
Just added arch-specific archives to the existing v3.1.3 release. I hope those fix the issue https://github.com/databus23/helm-diff/releases/tag/v3.1.3

damienpontifex:
Thanks @mumoshu to ensure I had a clean environment, I ran `helm plugin uninstall diff; sudo rm -r ~/Library/Caches/helm/plugins/https-github.com-databus23-helm-diff; helm plugin install https://github.com/databus23/helm-diff` and everything installed locally as expected :)

mumoshu:
@damienpontifex Glad to hear it worked! Thanks a lot for your confirmation ☺️

koenpunt:
[...]

PR #339, opened 2022-01-10
Fix install script

michaelsauter:
See https://github.com/databus23/helm-diff/issues/322#issuecomment-1008660815.

### DIFF
--- a/install-binary.sh
+++ b/install-binary.sh
@@ -19,7 +19,6 @@ fi
 [ -z "$HELM_BIN" ] && HELM_BIN=$(command -v helm)
 HELM_MAJOR_VERSION=$("${HELM_BIN}" version --client --short | awk -F '.' '{print $1}')
 
-HELM_HOME=$("${HELM_BIN}" home --debug=false)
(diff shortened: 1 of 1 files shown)


install-binary.sh
```diff
@@ -16,10 +16,9 @@ if type cygpath >/dev/null 2>&1; then
   HELM_PLUGIN_DIR="$(cygpath -u "${HELM_PLUGIN_DIR}")"
 fi
 
-[ -z "$HELM_BIN" ] && HELM_BIN=$(which helm)
+[ -z "$HELM_BIN" ] && HELM_BIN=$(command -v helm)
 HELM_MAJOR_VERSION=$("${HELM_BIN}" version --client --short | awk -F '.' '{print $1}')
 
-HELM_HOME=$("${HELM_BIN}" home --debug=false)
```
mumoshu, line 22: I think this was needed for Helm 2 but I hope no one's using helm 2 these days

mumoshu (approved):
Thanks for your contribution!

PR #340, opened 2022-01-10
helm-diff doesn't include values from stdin for helm3 version

ruslanloman:
Hi,
Unfortunately previous [PR](https://github.com/databus23/helm-diff/pull/272) was close as state, but this fix is still needed, so I've created a new one. Please review it once you have time. thanks!

There is an issue in helm-diff for helm3 version, it doesn't include values from stdin. For helm2 it works as expected.

```
$ cat values-10.yaml                                                               
lolkek:                                                                            
  helloword: test                                                                  
  new:                                                                             
    ohoho:                                                                         
      values-10: lolodddd10 
```
                                          
Current behavior:
```                                
$ cat values-10.yaml | helm3 diff upgrade test_helm . --allow-unreleased --values -
                                                                         
+ # Source: test_helm/templates/configMap.yaml                                     
+ apiVersion: v1                                                                   
+ kind: ConfigMap                                                                  
+ metadata:                                                                        
+   name: game-demo                                                                
+ data:                                                                            
+   user-interface.properties: |                                                   
+     null         
```
Expected behavior with applied fix:

```
$ cat values-10.yaml | helm3 diff upgrade test_helm . --allow-unreleased --values -
                                                                                 
+ # Source: test_helm/templates/configMap.yaml                                     
+ apiVersion: v1                                                                   
+ kind: ConfigMap                                                                  
+ metadata:                                                                        
+   name: game-demo                                                                
+ data:                                                                            
+   user-interface.properties: |                                                   
+     ohoho:
```

### DIFF
--- a/cmd/helm3.go
+++ b/cmd/helm3.go
@@ -120,7 +120,31 @@ func (d *diffCmd) template(isUpgrade bool) ([]byte, error) {
 		flags = append(flags, "--set-string", stringValue)
 	}
 	for _, valueFile := range d.valueFiles {
-		flags = append(flags, "--values", valueFile)
+		if strings.TrimSpace(valueFile) == "-" {
+			bytes, err := ioutil.ReadAll(os.Stdin)
+			if err != nil {
+				return nil, err
+			}
+
+			tmpfile, err := ioutil.TempFile("", "helm-diff-stdin-values")
+			if err != nil {
+				return nil, err
+			}
+			defer os.Remove(tmpfile.Name())
+
+			if _, err := tmpfile.Write(bytes); err != nil {
+				tmpfile.Close()
+				return nil, err
+			}
+
+			if err := tmpfile.Close(); err != nil {
+				return nil, err
+			}
+
+			flags = append(flags, "--values", tmpfile.Name())
+		} else {
+			flags = append(flags, "--values", valueFile)
+		}
 	}
 	for _, fileValue := range d.fileValues {
 		flags = append(flags, "--set-file", fileValue)


cmd/helm3.go
```diff
@@ -120,7 +120,26 @@ func (d *diffCmd) template(isUpgrade bool) ([]byte, error) {
...
 	}
 	for _, valueFile := range d.valueFiles {
-		flags = append(flags, "--values", valueFile)
+		if strings.TrimSpace(valueFile) == "-" {
+			var bytes []byte
+			var err error
+
+			bytes, err = ioutil.ReadAll(os.Stdin)
+
+			tmpfile, err := ioutil.TempFile("", "stdin-values")
```
mumoshu, line 129:
```suggestion
			tmpfile, err := ioutil.TempFile("", "helm-diff-stdin-values")
```

Perhaps this will better communicate that it's coming from helm-diff?

ruslanloman: Updated. thanks!

cmd/helm3.go
```diff
@@ -120,7 +120,26 @@ func (d *diffCmd) template(isUpgrade bool) ([]byte, error) {
 		flags = append(flags, "--set-string", stringValue)
 	}
 	for _, valueFile := range d.valueFiles {
-		flags = append(flags, "--values", valueFile)
+		if strings.TrimSpace(valueFile) == "-" {
+			var bytes []byte
+			var err error
+
+			bytes, err = ioutil.ReadAll(os.Stdin)
```
mumoshu, line 127:
```suggestion
			bytes, err := ioutil.ReadAll(os.Stdin)
			if err != nil {
				return nil, err
			}
```

Sry I missed it in my first iteration but you seem have missed handling the error and you don't need to separate var declarations as you don't mutate the vars `bytes` and `err` in the scope.

ruslanloman:
@mumoshu applied your changes. thanks!

cmd/helm3.go
```diff
@@ -120,7 +120,26 @@ func (d *diffCmd) template(isUpgrade bool) ([]byte, error) {
...
+
+			tmpfile, err := ioutil.TempFile("", "helm-diff-stdin-values")
+			if err != nil {
+				return nil, err
+			}
+			defer os.Remove(tmpfile.Name())
+
+			if _, err := tmpfile.Write(bytes); err != nil {
+				return nil, err
+			}
```
mumoshu, line 137: Did this change actually work for you?

I'm afraid we're missing `tmpfile.Close()` here, that I think is required to ensure writes are persisted. We should call `Close()` regardless of err is nil or not so wrapping it into another function call would be easier to read.

See `writeExistingValues` for your reference!

ruslanloman: Yes, it worked pretty well. I got your point regarding closing the tempfile. tbh i don't want to overcomplicate it by adding one more function, so just added one more if condition.

mumoshu: LGTM. Thanks for your effort!

mumoshu (approved):
I appreciate your contribution ☺️

PR #343, opened 2022-01-10
--color or HELM_DIFF_COLOR=true to force color

### LINKED ISSUE #344: Ability to force color
It turned out to be crucial when calling helm-diff from another command.

See https://github.com/roboll/helmfile/issues/2043 for more context.

mumoshu:
Resolves #344


gferon:
Thanks a lot for this! Would it be possible to release v3.3.3 with this patch? I miss the 🌈 in the CI output.

PR #346, opened 2022-01-11
Fix lint errors in install-binary.sh

Sajfer:
Cleanup of install-binary.sh and fix lint errors in the new github action job.


mumoshu (approved):
LGTM. Thanks for fixing it @Sajfer ☺️

PR #360, opened 2022-01-22
Add Alpine Linux target to CI

### LINKED ISSUE #347: ci: Add more Linux distros to the test matrix of install-binary.sh
I recently saw it can easily break on untested distros due to various shell script portability issues.
Thanks to #335 by @jkroepke we now have tests for cygwin, windows, and ubuntu.

GitHub Actions doesn't provide busybox or alpine VMs (it shouldn't 😃 ) so the remaining task would be to add containerized tests for install-binary.sh on those two distros, so that the coverage is complete.

Sajfer:
I have added alpine as a install target in the test matrix.
Alpine also uses busybox, so not sure if it needed to add a separate busybox target or if this is enough.

Fixes #347

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,6 +46,7 @@ jobs:
     if: "!contains(github.event.head_commit.message, '[ci skip]')"
     needs: [build]
     runs-on: ${{ matrix.os }}
+    container: ${{ matrix.container }}
     continue-on-error: ${{ matrix.experimental }}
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
Awesome! Thank you so much for your contribution @Sajfer ☺️ 

> Alpine also uses busybox, so not sure if it needed to add a separate busybox target or if this is enough.

Same here. Probably this can be enough. Let's see how it goes!

PR #373, opened 2022-02-14
Resolve security vulnerabilities

### LINKED ISSUE #372: Security vulnerabilities in golang modules
image (2)

[CVE-2018-15664](https://github.com/AOSC-Dev/aosc-os-abbs/issues/1891)
Severity: High
Component: github.com/docker/docker
Fix version: 18.09.7-rc1, 19.03.0-rc2

[CVE-2021-43565](https://github.com/golang/go/issues/49932)
Severity: High
Component: golang.org/x/crypto
Fix version: 0.0.0-20211202192323-5770296d904e

CVE-2021-32690
Severity: High
Component: k8s.io/helm
Fix version: 3.6.1

petrovicboban:
Fixes #372


go.mod
```diff
@@ -25,7 +25,7 @@ require (
 	k8s.io/apimachinery v0.21.0
 	k8s.io/cli-runtime v0.21.0
 	k8s.io/client-go v0.21.0
-	k8s.io/helm v2.16.12+incompatible
+	k8s.io/helm/v3 v3.6.1
```
mumoshu, line 28: This will never work as we still need the k8s.io/helm pkg for helm v2 support and `k8s.io/helm/v3` is for helm 3.

petrovicboban: How long do you plan to support helm v2?

z0rc:
Helm v2 is deprecated for more than a year per https://helm.sh/blog/helm-v2-deprecation-timeline/, there are zero support for this version, even from security standpoint. What's the point to maintaining helm v2 support here? Users who still use helm v2 can use older version of helm-diff.

Can we please just drop helm v2 support already?

mumoshu:
> What's the point to maintaining helm v2 support here? Users who still use helm v2 can use older version of helm-diff.
>
> Can we please just drop helm v2 support already?

@z0rc Maybe. I don't personally use helm v2 but I still occasionally see some users using helm v2 and helm-diff with helmfile and I don't know what i should do for them.

Anyway, helm-diff doesn't seem to be affected by the cve of helm and if you use helm-diff with helm v3 you have even less change of being affected by that cve. https://github.com/databus23/helm-diff/issues/372#issuecomment-1039881970

mumoshu (approved):
LGTM. Thanks for the fix!

PR #378, opened 2022-02-28
Annotation to override object name for diff purposes

ksa-real:
This is useful for immutable secrets which are typically named as secret-base-name-<.data hash>. Previously old and new secrets were considere completely independent ConfigMap/Secret. With this annotation user can add "helm.sh/base-name: secret-base-name" annotation and compare old and new objects similar to git rename.

#378

### DIFF
--- a/manifest/parse.go
+++ b/manifest/parse.go
@@ -7,7 +7,7 @@ import (
 	"log"
 	"strings"
 
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 	"k8s.io/helm/pkg/proto/hapi/release"
 )
 
@@ -40,8 +40,13 @@ func (m metadata) String() string {
 	if len(sp) > 1 {
 		apiBase = strings.Join(sp[:len(sp)-1], "/")
 	}
-
-	return fmt.Sprintf("%s, %s, %s (%s)", m.Metadata.Namespace, m.Metadata.Name, m.Kind, apiBase)
+	name := m.Metadata.Name
+	if a := m.Metadata.Annotations; a != nil {
+		if baseName, ok := a["helm-diff/base-name"]; ok {
+			name = baseName
+		}
+	}
+	return fmt.Sprintf("%s, %s, %s (%s)", m.Metadata.Namespace, name, m.Kind, apiBase)
 }
 
 func scanYamlSpecs(data []byte, atEOF bool) (advance int, token []byte, err error) {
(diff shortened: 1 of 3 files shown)


ksa-real:
@mumoshu Can you please take a look?

manifest/parse.go
```diff
@@ -40,8 +40,13 @@ func (m metadata) String() string {
 	if len(sp) > 1 {
 		apiBase = strings.Join(sp[:len(sp)-1], "/")
 	}
-
-	return fmt.Sprintf("%s, %s, %s (%s)", m.Metadata.Namespace, m.Metadata.Name, m.Kind, apiBase)
+	name := m.Metadata.Name
+	if a := m.Metadata.Annotations; a != nil {
+		if baseName, ok := a["helm.sh/base-name"]; ok {
```
mumoshu, line 45: Can we instead use anything else that communicates better about the fact helm-diff is independent from the helm project?(whereas `helm.sh/` prefix makes it look like it's from the upstream project

```suggestion
		if baseName, ok := a["helm-diff-base-name"]; ok {
```

manifest/parse.go
```diff
@@ -40,8 +40,13 @@ func (m metadata) String() string {
...
+			name = baseName
+		}
+	}
+	return fmt.Sprintf("%s, %s, %s (%s)", m.Metadata.Namespace, name, m.Kind, apiBase)
```
mumoshu, line 49: Does it still show the diff on `metadata.name`? Then it looks all good! If you could also add a test case to show how the output looks like when you diffed two secrets or configmaps, that would be great.

manifest/parse.go
ksa-real: How about "helm-diff/base-name" ?

manifest/parse.go
ksa-real: Yes, it does! I'll find some time in the near future will add the test.

ksa-real: And done.

mumoshu (approved):
LGTM. Thanks a lot for your contribution @ksa-real!

ksa-real:
Would you merge it?

mumoshu:
@ksa-real Yeah. I was just waiting for CI to pass.

ksa-real:
Do you do releases? It seems your release script is somewhat broken.

mumoshu:
@ksa-real I don't do releases so often so if you need to use it real quick, you'd better build and install locally by running `make install`

PR #379, opened 2022-03-01
diff/report: fix key regexp

eonpatapon:
The current regexp doesn't match all possible apis, for example
`cert-manager.io`.

As there is a closing `)` we can simply take everything before it.

Also resource names can have . in their name.

### DIFF
--- a/diff/report.go
+++ b/diff/report.go
@@ -191,7 +191,7 @@ func setupTemplateReport(r *Report) {
 // report with template output will only have access to ReportTemplateSpec.
 // This function reverts parsedMetadata.String()
 func (t *ReportTemplateSpec) loadFromKey(key string) error {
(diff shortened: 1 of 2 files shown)


eonpatapon:
https://github.com/databus23/helm-diff/pull/261
https://github.com/databus23/helm-diff/pull/290

mumoshu:
@eonpatapon Hey! Thanks a lot for your contribution. I'd appreciate it if you could add some unit tests to cover several possible and varying label keys, and a few impossible/invalid label keys.

eonpatapon:
@mumoshu I've added a test with some valid resources (some of which were problematic)

mumoshu (approved):
LGTM. Thank you so much for fixing it @eonpatapon!

PR #386, opened 2022-04-24
chore: Added support for setting --kube-version

semoac:
Support for `--kube-version` was ported to Helm 3 on https://github.com/helm/helm/pull/9040.

I'm using `helm-diff` as a mechanism to determine if a PR will cause changes on installed releases. Then, depending on the result, I trigger pipelines dynamically targeting only the affected releases. Currently, not all the clusters share the same Kubernetes version, so being able to use this Helm 3 feature on Helm diff will improve a lot my workflow.


mumoshu (approved):
LGTM. Thanks for your contribution @semoac!

PR #387, opened 2022-04-30
Added bzip2 to build docker. Updated golang image and github-release.

ksa-real:
Golang image was missing bzip2 so github-release binary didn't install
and release docker didn't work.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ docker-run-release: export pkg=/go/src/github.com/databus23/helm-diff
 docker-run-release:
 	git checkout master
 	git push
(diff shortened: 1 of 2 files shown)


ksa-real:
@mumoshu I have temporary released in my fork: https://github.com/ksa-real/helm-diff/releases (so team members can use the needed feature easier). The release docker couldn't find bzip2, which is pretty strange as you obviously had it somewhere. Please, let me know if I'm doing something incorrectly.

mumoshu (approved):
LGTM. Thanks!

PR #393, opened 2022-06-19
feat(secrets): allow diffing stringData sections of secrets as well as data sections

### LINKED ISSUE #302: Input stringData gets stripped completely
It appears as though the redactSecrets method is using `stringData` as a behind-the-scenes crutch to print a redacted `data` diff: https://github.com/databus23/helm-diff/blob/818e596fa6e9cb928985043c1b0a96b9b4a02299/diff/diff.go#L94-L129

The issue is that if the chart being diffed uses `stringData` then all those fields are completely hidden from the diff output. For example, this template:

```yaml
---
apiVersion: v1
kind: Secret
metadata:
  name: {{ $config.name }}
type: Opaque
stringData:
  kongCredType: acl
  group: inbound-traffic
  otherField: |
    Lorem ipsum dolor sit amet, consectetur adipiscing elit.
    Sed sed felis id ex ultricies tempor.
```

shows as this in helm-diff:

```diff
my-namespace, my-name, Secret (v1) has been added:
+ # Source: secret.yaml
+ apiVersion: v1
+ kind: Secret
+ metadata:
+   name: my-name
+ type: Opaque
```

This effectively defeats the diffing for any secret values that do not need to be redacted.

philomory:
Fixes #302.

If the same key appears in both the `data` and `stringData` sections of the same secret, the one in the `stringData` section is the one considered, because [that is how Kubernetes itself behaves](https://kubernetes.io/docs/concepts/configuration/secret/#restriction-names-data).

### DIFF
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -2,6 +2,7 @@ package diff
 
 import (
 	"bytes"
+	"encoding/base64"
 	"fmt"
 	"io"
 	"math"
@@ -163,10 +164,28 @@ func redactSecrets(old, new *manifest.MappingResult) {
 		if err := yaml.NewYAMLToJSONDecoder(bytes.NewBufferString(old.Content)).Decode(&oldSecret); err != nil {
 			old.Content = fmt.Sprintf("Error parsing old secret: %s", err)
 		}
+		if oldSecret.Data == nil {
+			oldSecret.Data = make(map[string][]byte, len(oldSecret.StringData))
+		}
+		for k, v := range oldSecret.StringData {
+			data := []byte(v)
+			dst := make([]byte, base64.StdEncoding.EncodedLen(len(data)))
+			base64.StdEncoding.Encode(dst, data)
+			oldSecret.Data[k] = dst
+		}
 	}
 	if new != nil {
 		if err := yaml.NewYAMLToJSONDecoder(bytes.NewBufferString(new.Content)).Decode(&newSecret); err != nil {
 			new.Content = fmt.Sprintf("Error parsing new secret: %s", err)
+		}
+		if newSecret.Data == nil {
+			newSecret.Data = make(map[string][]byte, len(newSecret.StringData))
+		}
+		for k, v := range newSecret.StringData {
+			data := []byte(v)
+			dst := make([]byte, base64.StdEncoding.EncodedLen(len(data)))
+			base64.StdEncoding.Encode(dst, data)
+			newSecret.Data[k] = dst
 		}
 	}
 	if old != nil {
--- a/install-binary.sh
+++ b/install-binary.sh
@@ -3,7 +3,7 @@
 # Shamelessly copied from https://github.com/technosophos/helm-template
 
 PROJECT_NAME="helm-diff"
-PROJECT_GH="philomory/$PROJECT_NAME"
+PROJECT_GH="databus23/$PROJECT_NAME"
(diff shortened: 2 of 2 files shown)


install-binary.sh
```diff
@@ -3,7 +3,7 @@
 # Shamelessly copied from https://github.com/technosophos/helm-template
 
 PROJECT_NAME="helm-diff"
-PROJECT_GH="databus23/$PROJECT_NAME"
+PROJECT_GH="philomory/$PROJECT_NAME"
```
mumoshu, line 6: Perhaps this is your temporary change on the script for local testing, which isn't intended to be included in this PR?

diff/diff.go
```diff
@@ -163,11 +164,29 @@ func redactSecrets(old, new *manifest.MappingResult) {
...
+		}
 	}
 	if new != nil {
 		if err := yaml.NewYAMLToJSONDecoder(bytes.NewBufferString(new.Content)).Decode(&newSecret); err != nil {
 			new.Content = fmt.Sprintf("Error parsing new secret: %s", err)
 		}
+		if newSecret.Data == nil {
+			newSecret.Data = make(map[string][]byte, len(newSecret.StringData))
+		}
+		for k, v := range newSecret.StringData {
```
mumoshu, line 184: I might be confused, but shouldn't we reset the whole `newSecret.Data` before merging in StringData entries? Otherwise, it might get redundant keys when you have different set of keys for data and stringData respectively?

In other words, what if it looked like:

```
data:
  foo: <BASE64 encoded value "FOO">
  bar: <BASE64 encoded value "BAR1">
stringData:
  bar: BAR2
  baz: BAZ
```

Should this function output:

```
stringData:
  foo: FOO
  bar: BAR2
  baz: BAZ
```

or

```
stringData:
  bar: BAR2
  baz: BAZ
```

?

install-binary.sh
philomory: Oh, whoops, sorry, that wasn't supposed to be included in the PR, hahaha

diff/diff.go
philomory: This is to handle the case where the incoming YAML manifest looks as follows:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: example
data:
  foo: <encoded:FOO>
  bar: <encoded:BAR>
stringData:
  bar: BAR2
  baz: BAZ
```

In such a case, per the Kubernetes documentation, the effective value of such a manifest is as follows:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: example
data:
  foo: <encoded:FOO>
  bar: <encoded:BAR2>
  baz: <encoded:BAZ>
```
... or, equivalently:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: example
stringData:
  foo: FOO
  bar: BAR2
  baz: BAZ
```

philomory: @mumoshu Specifically, reference the following from the [Kubernetes documentation](https://kubernetes.io/docs/concepts/configuration/secret/#restriction-names-data):
> All key-value pairs in the stringData field are internally merged into the data field. If a key appears in both the data and the stringData field, the value specified in the stringData field takes precedence.

philomory: @mumoshu Were there any other questions you had about this?

databus23:
Superseded by #407

PR #407, opened 2022-09-22
diff: fix secret redaction for secrets with stringData

### LINKED ISSUE #302: Input stringData gets stripped completely

majewsky:
Fixes #302.

This is an improved version of #393, with the following changes:

- The rewrite from `stringData` to `data` does not do an incorrect second Base64 encoding (the Base64 encoding is handled within k8s/client-go during YAML marshalling). For reference, [this is the equivalent code in the k8s apiserver](https://github.com/kubernetes/kubernetes/blob/2dba4034f8424574c42cebfce91664c4b7e08c61/pkg/apis/core/v1/conversion.go#L395-L403).
- Tests have been added that demonstrate diffing of redacted secrets with either `data` or `stringData` sections.

### DIFF
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -163,12 +163,29 @@ func redactSecrets(old, new *manifest.MappingResult) {
 		if err := yaml.NewYAMLToJSONDecoder(bytes.NewBufferString(old.Content)).Decode(&oldSecret); err != nil {
 			old.Content = fmt.Sprintf("Error parsing old secret: %s", err)
 		}
+		//if we have a Secret containing `stringData`, apply the same
+		//transformation that the apiserver would do with it (this protects
+		//stringData keys from being overwritten down below)
+		if len(oldSecret.StringData) > 0 && oldSecret.Data == nil {
+			oldSecret.Data = make(map[string][]byte, len(oldSecret.StringData))
+		}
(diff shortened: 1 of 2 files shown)


databus23 (approved):

databus23:
Thanks!

PR #410, opened 2022-09-26
fix(parser): don't always skip the first YAML document

### LINKED ISSUE #409: First existing manifest gets dropped
When using flux, the manifests don't contain a leading `---`.

Therefore https://github.com/databus23/helm-diff/blob/master/manifest/parse.go#L97-L98 leads to the first manifest being dropped.

I would propose to just delete these lines as https://github.com/databus23/helm-diff/blob/master/manifest/parse.go#L104-106 already filters out empty YAML documents

cwrau:
Just keep skipping empty documents

Closes #409


cwrau:
Heyho, can I do something to get this PR merged? 😊

cwrau:
I don't know why this is getting closed 😅

PR #416, opened 2022-10-14
feature(output): add environment variable for diff output format

cwrau:
Allow setting the output format via an environment variable


databus23 (approved):

databus23:
Thanks!

PR #426, opened 2022-12-12
Update README with actual usage output

max-len:
Not sure where the whitespaces in line 122 are coming from.


mumoshu (approved):
LGTM. Thanks for your patience and contribution @max-len!

PR #439, opened 2023-02-21
bump go from 1.18 to 1.19 and bump to helm 3.11.3

tbugfinder:


tbugfinder:
Not sure about the current process. @mumoshu would you review?

main.go
```diff
@@ -8,7 +8,6 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/exec"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
-	_ "k8s.io/client-go/plugin/pkg/client/auth/openstack"
```
mumoshu, line 11: @tbugfinder Hey! Thanks for your contribution. This looks perfect except this change.

Did you remove this due to that the OpenStack auth plugin isn't included in the upstream, by any chance? If so, do we have an alternative? Without an alternative, we might end up dropping support for OpenStack users.

tbugfinder: Yes, it was removed in upstream.

https://github.com/kubernetes/client-go/commit/98e81a77843e9ef2ad5c98a6919e450028610631

mumoshu: Thanks! So we'd need to add the external cloud provider explicitly to bring back the removed functionality here, right?

Context: https://github.com/kubernetes/kubernetes/pull/67782

mumoshu: Ah okay so they moved the client auth to the exec credentials and we already import `https://github.com/kubernetes/client-go/tree/master/plugin/pkg/client/auth/exec` above so... this looks good overall!

tbugfinder:
Merge conflicts were resolved and go.mod/go.sum updated.

mumoshu:
@tbugfinder Thanks for rebasing!

> now ending up in:

We probably need to bump k8s.io deps to v0.26.x as apparently those are required by helm 3.11.3

https://github.com/helm/helm/blob/66a969e7cc08af2377d055f4e6283c33ee84be33/go.mod#L38-L43

tbugfinder:
I've pushed a less aggressive update to go.mod :-D

tbugfinder:
@mumoshu could the workflow run be approved?

tehseenshahab:
@mumoshu Could you please merge this PR? Thanks

mumoshu (approved):
LGTM. Thanks a lot for your patience and contribution @tbugfinder!!

PR #454, opened 2023-04-24
Remove support for helm2

databus23:
I suggest to remove support for helm2 from helm diff as its EOL for years now.
I'm leaving this as PR for a while to collect some feedback if anybody is still using helm 2.


mumoshu (approved):
LGTM. Super excited to finally drop the v2 support! It has been a long journey!

PR #462, opened 2023-06-04
Add support for --insecure-skip-tls-verify

mumoshu:
I believe helm-diff should eventually support all the helm-template flags including connection-related ones, so that helm-diff can be used in place where helm-template or helm-upgrade-dry-run to provide diff/dry-run functionality to existing script/program/etc.

This adds support for --insecure-skip-tls which is used to disable server verification on chart download.


yxxhero (approved):

mumoshu:
Manually tested by locally running `HELM_BIN=$(which helm) ./helm-diff upgrade  --allow-unreleased --insecure-skip-tls-verify foo $CHART` via debugger!

PR #468, opened 2023-06-07
feat: Allow suppressing annotations and labels

### LINKED ISSUE #390: Secret: kubectl.kubernetes.io/last-applied-configuration not redacted
Hi, I have a case where a manually modified secret contains a `kubectl.kubernetes.io/last-applied-configuration` label with three-way-merge, containing the whole JSON of the previous secret version. The label is unfortunately printed out unredacted through helm diff. Maybe the plugin should handle this.

antoinedeschenes:
Allows suppressing annotations and labels in the diff output.

Fixes #390.

Useful to ignore changes that might not make an helm upgrade worthwhile, such as  chart version labels.
Also allows hiding labels containing sensitive information, such as `kubectl.kubernetes.io/last-applied-configuration` on secrets when using three-way-merge.

It currently doesn't handle pod templates in deployments, etc. which could still be very useful (to hide checksum annotation changes, for example).

### DIFF
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -11,6 +11,8 @@ import (
 	"github.com/aryann/difflib"
 	"github.com/mgutz/ansi"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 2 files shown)


diff/diff.go
```diff
@@ -101,6 +105,10 @@ func contentSearch(report *Report, possiblyRemoved []string, oldIndex map[string
 				redactSecrets(oldContent, newContent)
 			}
 
+			if len(options.SuppressedAnnotations) > 0 || len(options.SuppressedLabels) > 0 {
+				suppressAnnotationsAndLabels(oldContent, newContent, options)
```
mumoshu, line 109: Presuming all `diffMappingResults` calls need to be preceded by the new function calls, can we instead put this code inside `diffMappingResults` for maintainability, changing the third parameter of it to the `options` instead of a bool? 🐱

antoinedeschenes: Sorry for the late answer, I have running code also handling nested labels/annotations in pod templates, but I think it quickly gets too specialized, having to handle all cases one by one (CronJob, Job, Deployment, StatefulSet, etc.)

I think having a render script hook before diff could make sense, that would make it possible to run a custom script deleting/sanitizing required fields. 

I know there are post-renderer flags passed to helm, but these are only applied to the target helm template command.

WDYT?

yxxhero:
@antoinedeschenes any updates?

yxxhero:
https://github.com/databus23/helm-diff/pull/475

PR #471, opened 2023-06-09
add unittest cover

yxxhero:
add unittest cover

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ build: lint
 
 .PHONY: test
 test:
-	go test -v ./... -coverprofile cover.out -race -p=1
+	go test -v ./... -coverprofile cover.out -race
 	go tool cover -func cover.out
 
 .PHONY: bootstrap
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -39,7 +39,8 @@ build: lint
 
 .PHONY: test
 test:
-	go test -v ./...
+	go test -v ./... -coverprofile cover.out -race -p=1
```
mumoshu, line 42: @yxxhero Why do we need -p=1?

yxxhero: I don't know this flag. it comes from the internet. if you think that we don't need this. I will remove that. @mumoshu

mumoshu: Go ahead!
On a side note, we may find `-covermode=atomic` useful once we start using `t.Parallel()`.
However we don't use Parallel today so we could just remove `-p=1` for now.

yxxhero: got it.

PR #480, opened 2023-07-03
feat: Transparent release process

bonddim:
- Added goreleaser config
- Add release workflow

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,0 +1,51 @@
+---
+# Make sure to check the documentation at https://goreleaser.com
+before:
+  hooks:
+    # Keep mod files clean
+    - go mod tidy -v
+
+builds:
+  - main: main.go
+    binary: bin/diff
+    env:
+      - CGO_ENABLED=0
+    flags: -trimpath
+    ldflags: -s -w -X {{ .ModulePath }}/cmd.Version={{ .Version }}
+    targets:
+      # helm supported targets
+      # https://github.com/helm/helm/blob/main/Makefile#L4C1-L4C1
+      - darwin_amd64
+      - darwin_arm64
+      - linux_386
+      - linux_amd64
+      - linux_arm
+      - linux_arm64
+      - linux_ppc64le
+      - linux_s390x
+      - windows_amd64
+      # additional targets
+      - freebsd_amd64
+
+archives:
+  - format: tgz
+    name_template: '{{ .ProjectName }}-{{ tolower .Os }}-{{ .Arch }}'
+    wrap_in_directory: diff
+    files:
+      - README.md
+      - LICENSE
+      - plugin.yaml
+
+checksum:
+  name_template: 'checksums.txt'
+
+snapshot:
+  name_template: '{{ incpatch .Version }}-{{ .ShortCommit }}'
+
+changelog:
+  sort: asc
+  filters:
+    exclude:
+      - '^chore:'
+      - '^docs:'
+      - '^test:'
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -8,4 +8,3 @@ fi
 
 git tag $1
 git push origin $1
-gh release create $1 --draft --generate-notes --title "$1" release/*.tgz
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,4 +1,4 @@
-name: release
+name: Release
 
 on:
   push:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,32 @@
+name: release
+
+on:
+  push:
+    tags:
+      - '*'
+
+permissions:
+  contents: write
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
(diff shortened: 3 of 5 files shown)


mumoshu:
@bonddim Thanks! Just wondering, how did you test this?! If you've already run goreleaser using this config on your repo, I'd love it if you could share it for review 🙏

bonddim:
@mumoshu Sure,
Published [release](https://github.com/bonddim/helm-diff/releases/tag/v3.8.2)
Workflow [run](https://github.com/bonddim/helm-diff/actions/runs/5653796404/job/15315599563).

Note that this was first tag in my forked repo, that's why goreleaser fill changelog with a lot of commits.

.goreleaser.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - linux_ppc64le
+      - linux_s390x
+      - windows_amd64
+      # additional targets
+      - freebsd_amd64
+
+archives:
+  - format: tgz
+    name_template: '{{ .ProjectName }}-{{ tolower .Os }}-{{ .Arch }}'
+    wrap_in_directory: diff
```
mumoshu, line 33: I think we don't need this wrapping, because our installation script doesn't expect it.

.goreleaser.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - LICENSE
+      - plugin.yaml
+
+checksum:
+  name_template: 'checksums.txt'
+
+snapshot:
+  name_template: '{{ incpatch .Version }}-{{ .ShortCommit }}'
+
+changelog:
```
mumoshu, line 45: I missed this in my PR so let me check later!

.goreleaser.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - linux_arm64
+      - linux_ppc64le
+      - linux_s390x
+      - windows_amd64
+      # additional targets
+      - freebsd_amd64
+
+archives:
+  - format: tgz
+    name_template: '{{ .ProjectName }}-{{ tolower .Os }}-{{ .Arch }}'
```
mumoshu, line 32: It turned out we had to use `macos` instead of `darwin` in the archive name.

https://github.com/databus23/helm-diff/pull/517/files#diff-42e26dc67aed8aa3edb2472b4403288c1699fb6dc47419b9a475f0f224fe4689R7

.goreleaser.yaml
```diff
@@ -0,0 +1,51 @@
...
+    flags: -trimpath
+    ldflags: -s -w -X {{ .ModulePath }}/cmd.Version={{ .Version }}
+    targets:
+      # helm supported targets
+      # https://github.com/helm/helm/blob/main/Makefile#L4C1-L4C1
+      - darwin_amd64
+      - darwin_arm64
+      - linux_386
+      - linux_amd64
+      - linux_arm
```
mumoshu, line 22: If I'm not missing anything, we can safely omit `386`, `arm`, `ppc64le` and `s390x` "yet" as we don't publish binaries for those archs today.

mumoshu (commented):
Thank you for your contribution!
I wanted to introduce MVP of release automation and do incremental enhancements on top of that, and #517 is my first attempt.
I've commented on this PR to note differences. If you are still around, I'd love your confirmation and feedback!

.goreleaser.yaml
bonddim: All files in latest release are wrapped into `diff` directory, and installation script uses it :
https://github.com/databus23/helm-diff/blob/f5cb54251336165ce792d8d7fb43329306a1a9c1/install-binary.sh#L126

.goreleaser.yaml
bonddim: Doesn't required for releases, but I like this snapshot versioning

.goreleaser.yaml
bonddim: This is correct - I missed this.

.goreleaser.yaml
bonddim: Just included all architectures supported by Helm. 
Yes, its safe to omit them, but why to not include it now?

scripts/release.sh
```diff
@@ -8,4 +8,3 @@ fi
 
 git tag $1
 git push origin $1
-gh release create $1 --draft --generate-notes --title "$1" release/*.tgz
```
mumoshu, line 11: Good catch! Today, the release process looks like the below:

- Add a commit that updates plugin.yaml to have a new version
- Run this release.sh

After this and #517, the process would look like:

- Add a commit that updates plugin.yaml to have a new version
- Create a semver tag along with the release on the GitHub web UI
- The workflow gets triggered on GHA. goreleaser does the rest.

I guess we'd rather prefer the below:

- Add a commit that updates plugin.yaml to have a new version
- A GHA workflow reads the new plugin.yaml version value and tags the commit with the version
- The release workflow gets triggered and goreleaser does the rest.

Maybe we can modify this release.sh to do the second bullet point?
Basically, we can deduce `$1` here from `plugin.yml` by reading the version field using eg. `yq`? 🤔

.goreleaser.yaml
mumoshu: You are absolutely correct @bonddim! Thank you very much for the confirmation and the correction. You saved the day ❤️

Issue #163, opened 2019-11-04
Suppress example


ragarcia26:
Upgrade diff's can be noisy when you have something like a chart label changing every release which will output the whole file. I was looking at `--suppress spec.template.metadata.labels.helm.sh/chart` to help with that, but it doesn't appear to work and can't find any actual usage examples.

schollii:
same issue, basically if `helm diff upgrade RELEASE_NAME REPO/CHART_NAME --verison CHART_VERSION --namespace NSNAME` shows a diff on a.b.c and e.f.g, I thought that the same command with `--suppress a.b.c,e.f.g` would not show these diff, but it does. I think there is something missing from docs because these property names can appear in several kubernetes specs of the chart (say, metadata.labels.chart is in both service and deployment and pod), what if I only want one of them, or all of them, to be ignored?

schollii:
So after looking at the code, the `--suppress` is just poorly documented: you can give it a *kind* of object, say Deployment or Service, to suppress. E.g. `helm diff upgrade RELEASE_NAME REPO/CHART_NAME --version CHART_VERSION --namespace NSNAME --suppress Deployment,Service` will suppress all diffs in the k8s objects of `kind=Deployment` and `kind=Service`.

schollii:
Could there be a robust way to implement `--suppress-pattern`, like perhaps `--suppress-pattern *.metadata.chart.labels` would suppress. One problem is that the objects are not named, e.g. there could be multiple Deployment in the chart, how to select them? One way would be with a selector perhaps. This could get pretty heavy on the command line but to filter out diffs that you don't care about, you could write a yaml like this: 
```
ignore:
   selector: 
       metadata.chart.labels:
           - something1: some-value
           - something2: other-value
   diffs:
       - metadata.chart.labels
       - spec.template.spec.something: ...
```
Too complicated?

ronyjoy:
I couldn't make the helm diff suppress to work , can any one paste a working example.  I tried the following in the helm diff  but as soon as i add --suppress the command never finish it is stuck for ever 

--suppress Secret.data=[create_vhosts]

jcribeirooo:
It is still an issue. Can we re-open it?

PR #491, opened 2023-09-12
Fix description of diff suppress option

yktakaha4:
As also mentioned in the issue below, the current `--suppress` option only handles kind, but the vague description is causing confusion.
ref: https://github.com/databus23/helm-diff/issues/163

I think #468 is the fundamental solution, but first I would like to clarify the current specifications.


databus23 (approved):

yxxhero (approved):

mumoshu:
@yktakaha4 Awesome! Thanks for fixing it ☺️

PR #458, opened 2023-05-09
Use dry-run=server to enable lookups

MichaelMorrisEst:
Helm PR https://github.com/helm/helm/pull/9426 enables support for executing lookups during dry run. This PR is to make use of this new support in helm-diff. Backwards compatibility for older versions of helm is maintained by checking the helm version before setting the flag

Addresses issue: https://github.com/databus23/helm-diff/issues/449

Issue #449, opened 2023-04-11
Use new helm support for lookups in helm dry-run


MichaelMorrisEst:
[PR 9426](https://github.com/helm/helm/pull/9426) in helm will allow setting the value for the --dry-run flag to enable the execution of lookups towards the cluster.
I would like to submit a PR that will use this new option in order to solve https://github.com/databus23/helm-diff/issues/263

I am hopeful the PR in helm will be merged soon and would like to start working on the implementation in helm-diff and would appreciate any thoughts. In particular, is it ok to change the implementation to always use "--dry-run=server" (i.e. execute lookups) or do we need the existing behaviour by default and a new flag in helm-diff to enable using the new helm flag?

MichaelMorrisEst:
@databus23 any thoughts on how you would like to see this implemented?
i.e. is it ok to change the implementation to always use "--dry-run=server" (i.e. execute lookups) or do we need the existing behaviour by default and a new flag in helm-diff to enable using the new helm flag?

databus23:
If the outcome as identical to the current output with the benefit that the `lookup` function is evaluated as it will be during upgrade I'm all for always using `--dry-run=server`. But we need to maintain backward compatibility with older helm versions that don't support the new flag yet and fallback to the currently used command invocation.

dudicoco:
I'm all for the new flag as I've been waiting for it as well, however it should be added explicitly by the user due to the fact that this will make helm contact the k8s api server, which could be security concern for some users, especially when using a third party chart.

mumoshu:
The support in helm-diff is available since 3.9.0 and considered complete since 3.9.4 released today! Enjoy!

PR #499, opened 2023-10-11
fix: Make dry-run=server optional

mumoshu:
This intends to fix a potential security issue introduced via #458 before cutting the next helm-diff release.

Since #458 (unreleased), we had forced helm-diff to use `helm template --dry-run=server` for Helm 3.13 or greater.

I think this can create an unintended security hole, where any users, who can run helm-diff via CI or any automation with an arbitrary chart and values, is able to view cluster resources via helm template's `lookup` functions.

Previously this was impossible because `helm template` run by `helm diff` had no access to the `lookup` function. To fix this, we need to make `--dry-run=server` optional. And we do so by changing helm-diff's `--dry-run` flag to accept only only booleans but also `client` and `server`. The updated flag usage is `--dry-run[=[|true|false|client|server]]`.

See the updated README and the updated helm-diff help message for more details.

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -13,6 +13,7 @@ import (
 	jsoniterator "github.com/json-iterator/go"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/kube"
@@ -37,7 +38,8 @@ type diffCmd struct {
 	devel                    bool
 	disableValidation        bool
 	disableOpenAPIValidation bool
-	dryRun                   bool
+	dryRunMode               string
+	dryRunModeSpecified      bool
 	namespace                string // namespace to assume the release to be installed into. Defaults to the current kube config namespace.
 	valueFiles               valueFiles
 	values                   []string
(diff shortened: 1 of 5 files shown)


cmd/upgrade.go
```diff
@@ -67,6 +69,11 @@ func (d *diffCmd) isAllowUnreleased() bool {
 	return d.allowUnreleased || d.install
 }
 
+func (d *diffCmd) isRemoteAccessAllowed() bool {
```
yxxhero, line 72: adding some unit tests will be better.

mumoshu: @yxxhero Are you saying we should have a unit test for checking all the combinations of variables involved in L73?

yxxhero: yeah.

yxxhero (approved):

jBouyoud:
Any ETA for this PR? and new release ? 🥺 🙏

gdubicki:
Related to https://github.com/databus23/helm-diff/issues/449

yxxhero:
@mumoshu

cavedon:
It looks like the `-f` flag stopped working with this patch:
```
 % helm diff upgrade -f secrets.yaml -n map map .        
args after legacy dry-run parsing: [-f secrets.yaml map .]
Error: Failed to get release -f in namespace map: exit status 1: Error: unknown shorthand flag: 'f' in -f

```

mumoshu:
@cavedon Thank you very much for testing! I fixed it and added some tests to verify flag parsing is working.

jBouyoud:
@mumoshu @yxxhero Thanks very much for this PR. Have you now any ETA for a new release ? 
There are some incredible new features that I'd like to use (server dry-run, --suppress-output-line-regex option , and beyond) ;-) . I'll also glad to help if I can, let me know

PR #507, opened 2023-11-03
lint: Update depguard configuration

mjtrangoni:
Hi @databus23 @mumoshu,

This update the [depguard](https://github.com/OpenPeeDeeP/depguard) *golangci-lint* configuration to the latest state, and update golangi-lint to `1.55.1` as well.

At the same time, this configuration update pointed out the usage of `github.com/pkg/errors`. 

```console
    cmd/upgrade.go:14:2: import 'github.com/pkg/errors' is not allowed from list 'main': Should be replaced by standard lib errors package (depguard)
            "github.com/pkg/errors"
            ^
```

and I fixed it in a separated commit.

Thank you!

Mario


yxxhero:
@mjtrangoni please rebase this PR.

mjtrangoni:
@yxxhero done. Thank you!

yxxhero:
@mumoshu WDYT?

.golangci.yaml
```diff
@@ -138,14 +138,44 @@ linters-settings:
...
+  depguard:
+    # Rules to apply.
+    # Default: Only allow $gostd in all files.
+    rules:
+      # Name of a rule.
+      main:
+        # List of file globs that will match this list of settings to compare against.
+        # Default: $all
+        files:
+          - "!**/*_a _file.go"
```
mumoshu, line 150: @mjtrangoni Does this literally match everything except `**/_a _file.go`? Then, I guess this is basically the same as `$all`.. no?

mumoshu (commented):
Thanks for your patience! Just fixed the conflict and reviewed this.
Had only one comment so this can be merged as soon as that's answered/resolved!

.golangci.yaml
mjtrangoni: @mumoshu indeed it is like this, but the basic example from upstream use that funny regex. See [here](https://golangci-lint.run/usage/linters/#depguard). Should I change it back to `$all`?

yxxhero: Try to use all.

mumoshu: Thanks for confirming! Let's use `$all` for now and see how it goes.

```suggestion
          - $all
```

yxxhero (approved):

PR #517, opened 2024-01-10
Automate releases using GHA

mumoshu:
To publish a release, we currently have to run `make docker-run-release` locally after tagging a release, which is not a huge effort but a bit cumbersome because we do it only a few times a year and that is long enough to lose my memory :)

Starting this change, we use a GHA workflow and goreleaser to automatically build and publish binaries whenever a new semver tag is created, so that we do not need to run `make` anymore.

I crafted the goreleaser configuration so that it produces the same archives that we use and helm-diff's install.sh relies on.

For example, the produced archives has contents like the below:

- plugin.yaml
- README.md
- LICENSE
- bin/diff

`bin/diff` is especially important because that's the path we use in the plugin.yaml.

You can test this almost locally, by running `goreleaser --snapshot --clean`. Whenever you modify `.goreleaser.yml`, I'd recommend you running that for verification.

I bootstraped this from my other projects. But the end result turned out to be almost the same as @bonddim's awesome PR #480 submitted last July(!) so big applause to @bonddim for his awesome work!

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -24,7 +24,7 @@ jobs:
         name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: '1.20.12'
+          go-version-file: 'go.mod'
           cache: true
       -
         name: Run GoReleaser
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,37 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - '*'
+
+permissions:
+  contents: write
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,37 @@
...
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      -
+        name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.20.12'
```
bonddim, line 27: I'd rather use `go-version-file: 'go.mod'`
You don't need to update workflow in case of go patch version bump.

yxxhero: good idea.

mumoshu: Updated in [6ab9a0b](https://github.com/databus23/helm-diff/pull/517/commits/6ab9a0b7b277b254a436dc8ffda8dbc5cbc0754a). Thank you very much for the suggestion ☺️

yxxhero (approved):
LGTM

PR #519, opened 2024-01-13
fix: remove debug msg in normal mode

yxxhero:
```
+args after legacy dry-run parsing: [--allow-unreleased manifests /tmp/chartify3644874502/helmfile-tests/manifests --reset-values]
 ********************
 
 	Release was not present in Helm.  Diff will show entire contents as new.
FAIL: "helmfile diff" should be consistent
```

/cc @mumoshu


mumoshu (approved):
LGTM! Thanks for fixing it 👍

yxxhero:
@mumoshu should we release 3.9.1? WDYT?

yxxhero:
@mumoshu all passed.

Issue #537, opened 2024-01-19
HELM_DIFF_IGNORE_UNKNOWN_FLAGS seems to be ignored in v3.9.2


deiga:
Hello,

We have added `--force` to `helm diff upgrade` a few weeks ago and added `HELM_DIFF_IGNORE_UNKNOWN_FLAGS=true` to the environment.

Today our client auto-updated to `3.9.2` from `3.8.1` and we are getting this error in our CI. 

```
[2024-01-18T19:15:22.172Z] + helm diff -C7 upgrade --allow-unreleased --force --values helm/values-eu-qa.yaml --values helm/values-pr.yaml ...

[2024-01-18T19:15:22.172Z] Error: unknown flag: --force

[2024-01-18T19:15:22.172Z] Usage:

[2024-01-18T19:15:22.172Z]   diff upgrade [flags] [RELEASE] [CHART]

```

deiga:
We reverted back to `3.8.1` and it works again

yxxhero:
@deiga PR is welcome.

yxxhero:
https://github.com/databus23/helm-diff/pull/499

PR #538, opened 2024-01-19
fix: HELM_DIFF_IGNORE_UNKNOWN_FLAGS ignored issue

yxxhero:
fix: https://github.com/databus23/helm-diff/issues/537


yxxhero:
@mumoshu

cmd/upgrade.go
```diff
@@ -164,13 +165,18 @@ func newChartCommand() *cobra.Command {
 						" --dry-run=server enables the cluster access with helm-get and the lookup template function."
 				)
 
+				cmdFlags := cmd.Flags()
+
+				// see: https://github.com/databus23/helm-diff/issues/537
+				cmdFlags.ParseErrorsWhitelist.UnknownFlags = unknownFlags
```
mumoshu, line 171: Wow, good to know! Thanks..

So does this basically mean `DisableFlagParsing: true,` results in `UnknownFlags NOT set by cobra? (Just confirming for learning purpose

yxxhero: Yes, this will disrupt a lot of functionality. I think we only support -- dry run=something. @mumoshu WDYT?

mumoshu: @yxxhero Thanks! Re dry-run, it won't be a good idea to start dropping the backward-compatibility without any prenotices. We are still in the same major ver of helm-diff (3.x). Dropping the compatibility may break downstream user CI/CD pipelines and/or tools.

I think this justifies having more test coverage in future though 🤔

yxxhero: @mumoshu yeah. make sense.

mumoshu (approved):
@yxxhero Awesome work! Going to merge this and cut a new release asap.

yxxhero:
@mumoshu there are some other issue should be fixed. do you have any good iede for fixing issue that caused by DisableFlagParsing: true?

PR #544, opened 2024-01-26
Add enable dns flag

### LINKED ISSUE #447: Support helm flag --enable-dns
With helm 3.11.1, the --enabe-dns flag was introduced. Background for this was https://github.com/advisories/GHSA-pwcw-6f5g-gxf8. The PR was this https://github.com/helm/helm/commit/5abcf74227bfe8e5a3dbf105fe62e7b12deb58d2. 

Unfortunately, helm diff doesn't support this yet to my knowledge. Do we simply need to bump the helm dependency here?

```bash
helm diff upgrade --enable-dns test --values test.yaml --values
Error: unknown flag: --enable-dns
```

hesstobi:
Closes #447

This adds the `--enable-dns` option to the upgrade command. And enables DNS Lookup during rendering the templates. 

I didn't add any tests, because there are also no test for the other options. If tests are required, can you give my hints where to start.


mumoshu (approved):
LGTM. Thanks for raising the issue and submitting this PR!

Re; tests, I think that's not strictly required for this case right now. All we have to do here is, pass `enable-dns` as-is from helm-diff to downstream helm-template.

We could later extract the part of the `template` function that composes `flags`, `args`, `cmd` and write some unit test for that extracted function, though!
If you could submit follow-up PR(s) to do so, that would be awesome!

PR #547, opened 2024-01-27
Remake dry-run=server optional

### LINKED ISSUE #536: `--color` flag doesn't work anymore
In v3.9.2. `--color` doesn't force a coloured output anymore.

- In 2.8.1 a command such as `helm diff upgrade NAME CHART --color | cat` would show a coloured result.
- In 2.9.2 the output is monochrome.

If the output is direct to TTY, 2.9.2 produces a coloured output as well. It's the `--color` flag to force it to seem broken.

mumoshu:
This is a follow-up on #499 which redoes it without relying on cobra's DisableFlagParsing. DisableFlagParsing turned out to break far many things than I imagined!

Fixes #536

Ref #537 #538

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -13,7 +13,6 @@ import (
 	jsonpatch "github.com/evanphx/json-patch"
 	jsoniterator "github.com/json-iterator/go"
 	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/kube"
@@ -29,6 +28,10 @@ import (
 	"github.com/databus23/helm-diff/v3/manifest"
 )
 
+const (
+	dryRunNoOptDefVal = "client"
+)
+
 type diffCmd struct {
 	release                  string
 	chart                    string
@@ -38,8 +41,6 @@ type diffCmd struct {
 	devel                    bool
 	disableValidation        bool
 	disableOpenAPIValidation bool
-	dryRunMode               string
-	dryRunModeSpecified      bool
 	enableDNS                bool
 	namespace                string // namespace to assume the release to be installed into. Defaults to the current kube config namespace.
 	valueFiles               valueFiles
@@ -62,6 +63,14 @@ type diffCmd struct {
 	kubeVersion              string
 	useUpgradeDryRun         bool
 	diff.Options
+
+	// dryRunMode can take the following values:
+	// - "none": no dry run is performed
+	// - "client": dry run is performed without remote cluster access
+	// - "server": dry run is performed with remote cluster access
+	// - "true": same as "client"
+	// - "false": same as "none"
+	dryRunMode string
 }
 
 func (d *diffCmd) isAllowUnreleased() bool {
@@ -73,10 +82,9 @@ func (d *diffCmd) isAllowUnreleased() bool {
 
 // clusterAccessAllowed returns true if the diff command is allowed to access the cluster at some degree.
 //
-// helm-diff basically have 3 modes of operation:
-// 1. no cluster access at all when --dry-run, --dry-run=client, or --dry-run=true is specified.
-// 2. basic cluster access when --dry-run is not specified.
-// 3. extra cluster access when --dry-run=server is specified.
+// helm-diff basically have 2 modes of operation:
+// 1. without cluster access at all when --dry-run=true or --dry-run=client is specified.
+// 2. with cluster access when --dry-run is unspecified, false, or server.
 //
 // clusterAccessAllowed returns true when the mode is either 2 or 3.
 //
@@ -89,8 +97,7 @@ func (d *diffCmd) isAllowUnreleased() bool {
 //
 // See also https://github.com/helm/helm/pull/9426#discussion_r1181397259
 func (d *diffCmd) clusterAccessAllowed() bool {
-	clientOnly := (d.dryRunModeSpecified && d.dryRunMode == "") || d.dryRunMode == "true" || d.dryRunMode == "client"
-	return !clientOnly
+	return d.dryRunMode == "none" || d.dryRunMode == "false" || d.dryRunMode == "server"
 }
 
 const globalUsage = `Show a diff explaining what a helm upgrade would change.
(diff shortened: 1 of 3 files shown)


cmd/upgrade.go
```diff
@@ -293,6 +231,9 @@ func newChartCommand() *cobra.Command {
 	f.BoolVar(&diff.devel, "devel", false, "use development versions, too. Equivalent to version '>0.0.0-0'. If --version is set, this is ignored.")
 	f.BoolVar(&diff.disableValidation, "disable-validation", false, "disables rendered templates validation against the Kubernetes cluster you are currently pointing to. This is the same validation performed on an install")
 	f.BoolVar(&diff.disableOpenAPIValidation, "disable-openapi-validation", false, "disables rendered templates validation against the Kubernetes OpenAPI Schema")
+	f.StringVar(&diff.dryRunMode, "dry-run", "", "--dry-run, --dry-run=client, or --dry-run=true disables cluster access and show diff as if it was install. Implies --install, --reset-values, and --disable-validation."+
+		" --dry-run=server enables the cluster access with helm-get and the lookup template function.")
+	f.Lookup("dry-run").NoOptDefVal = "true"
```
mumoshu, line 236: So this was the biggest finding and improvement for me.
This is how helm kept the backward-compatibility on the flag when they introduced `dry-run=server` https://github.com/tapaskapadia/helm/blob/b7a2d47eca70e18b821b415efa5f47971c8a5302/cmd/helm/install.go#L166-L167

cmd/upgrade.go
```diff
@@ -153,71 +150,12 @@ func newChartCommand() *cobra.Command {
 			"# Read the flag usage below for more information on --context.",
 			"HELM_DIFF_OUTPUT_CONTEXT=5 helm diff upgrade my-release datadog/datadog",
 		}, "\n"),
+		Args: func(cmd *cobra.Command, args []string) error {
+			return checkArgsLength(len(args), "release name", "chart path")
+		},
```
mumoshu, line 155: We just moved this checkArgsLength from RunE to Args, as we used to do so before #499

cmd/upgrade.go
```diff
@@ -89,7 +87,7 @@ func (d *diffCmd) isAllowUnreleased() bool {
 //
 // See also https://github.com/helm/helm/pull/9426#discussion_r1181397259
 func (d *diffCmd) clusterAccessAllowed() bool {
-	clientOnly := (d.dryRunModeSpecified && d.dryRunMode == "") || d.dryRunMode == "true" || d.dryRunMode == "client"
+	clientOnly := d.dryRunMode == "true" || d.dryRunMode == "client"
```
mumoshu, line 90: We could do better, but the intention is; we ensure that `dryRunMode` ever takes any of `true`, `false`, `client` and `server` by doing `if diff.dryRunMode != "true" && diff.dryRunMode != "false" && diff.dryRunMode != "client" && diff.dryRunMode != "server" {` check in `cmd/upgrade.go`.

It's enough to check for `true` and `client` to say it's `clientOnly`.
If the user didn't told us to be `clientOnly`, it implies `clusterAccessAllowed`.

cmd/upgrade.go
```diff
@@ -293,6 +242,9 @@ func newChartCommand() *cobra.Command {
 	f.BoolVar(&diff.devel, "devel", false, "use development versions, too. Equivalent to version '>0.0.0-0'. If --version is set, this is ignored.")
 	f.BoolVar(&diff.disableValidation, "disable-validation", false, "disables rendered templates validation against the Kubernetes cluster you are currently pointing to. This is the same validation performed on an install")
 	f.BoolVar(&diff.disableOpenAPIValidation, "disable-openapi-validation", false, "disables rendered templates validation against the Kubernetes OpenAPI Schema")
+	f.StringVar(&diff.dryRunMode, "dry-run", "", "--dry-run, --dry-run=client, or --dry-run=true disables cluster access and show diff as if it was install. Implies --install, --reset-values, and --disable-validation."+
+		" --dry-run=server enables the cluster access with helm-get and the lookup template function.")
+	f.Lookup("dry-run").NoOptDefVal = dryRunNoOptDefVal
```
mumoshu, line 247: This was the most significant finding and improvement for me.
This is how helm kept the backward-compatibility on the flag when they introduced dry-run=server https://github.com/tapaskapadia/helm/blob/b7a2d47eca70e18b821b415efa5f47971c8a5302/cmd/helm/install.go#L166-L167

cmd/upgrade.go
```diff
@@ -153,71 +159,14 @@ func newChartCommand() *cobra.Command {
...
-				// cmd.Args = func(cmd *cobra.Command, args []string) error {
-				//     return checkArgsLength(len(args), "release name", "chart path")
-				// }
-				//
-				// Because it seems to take precedence over the PersistentPreRunE
-				if err := checkArgsLength(len(args), "release name", "chart path"); err != nil {
-					return err
-				}
+			if diff.dryRunMode == "" {
+				diff.dryRunMode = "none"
```
mumoshu, line 167: We can indeed use either `""` or `"none"` to denote that the dry run flag is unspecified.
However I intentionally chose `"none"` to make code consistent with helm. They use `"none"` to indicate unspecified dry-run too..

cmd/upgrade_test.go
```diff
@@ -9,39 +9,51 @@ func TestIsRemoteAccessAllowed(t *testing.T) {
...
 			},
 			expected: false,
 		},
+		{
+			name: "legacy explicit dry-run=false flag",
+			cmd: diffCmd{
+				dryRunMode: "false",
+			},
+			expected: true,
+		},
```
mumoshu, line 31: I missed adding the case for `false` in #499 so doing it here.

cmd/upgrade_test.go
```diff
@@ -9,39 +9,51 @@ func TestIsRemoteAccessAllowed(t *testing.T) {
 		expected bool
 	}{
 		{
-			name:     "no flags",
-			cmd:      diffCmd{},
+			name: "no flags",
+			cmd: diffCmd{
+				dryRunMode: "none",
```
mumoshu, line 14: As per https://github.com/databus23/helm-diff/pull/547/files#r1468284710

yxxhero:
@mumoshu test locally. all is ok.

yxxhero (approved):

Issue #503, opened 2023-10-26
helm 3.10.0 helm diff --insecure-skip-tls-verify Error: unknown flag: --insecure-skip-tls-verify


HuLaLaGa:

HuLaLaGa:
Helm: 3.10.0
Helm-Diff: 3.8.1
command: helm diff --insecure-skip-tls-verify
error: Error: unknown flag: --insecure-skip-tls-verify

tjanson:
I ran into this as well with Helm v3.13.1, helm-diff v3.8.1. Did this work with any previous version? (Then perhaps I could downgrade.)

tjanson:
I guess this is a bug in #462 ?

HuLaLaGa:
![image](https://github.com/databus23/helm-diff/assets/34408166/82e306d5-6ee6-4f85-a4e9-aa04b70d6069)

HuLaLaGa:
![image](https://github.com/databus23/helm-diff/assets/34408166/70089cb5-c6ab-4d64-a982-c041161f7979)

HuLaLaGa:
![image](https://github.com/databus23/helm-diff/assets/34408166/174b3509-cbb8-4eb2-9d88-711c8e89d19f)

yxxhero:
@tjanson PR is welcome.

yxxhero:
I can's reproduce this issue.

yxxhero:
see: https://github.com/helm/helm/pull/12856

PR #568, opened 2024-03-03
rename insecure-skip-tls-verify to helm-insecure-skip-tls-verify

yxxhero:
see: https://github.com/databus23/helm-diff/issues/503

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -253,7 +253,8 @@ func newChartCommand() *cobra.Command {
 	f.BoolVar(&diff.enableDNS, "enable-dns", false, "enable DNS lookups when rendering templates")
 	f.StringVar(&diff.postRenderer, "post-renderer", "", "the path to an executable to be used for post rendering. If it exists in $PATH, the binary will be used, otherwise it will try to look for the executable at the given path")
 	f.StringArrayVar(&diff.postRendererArgs, "post-renderer-args", []string{}, "an argument to the post-renderer (can specify multiple)")
-	f.BoolVar(&diff.insecureSkipTLSVerify, "insecure-skip-tls-verify", false, "skip tls certificate checks for the chart download")
+	// see https://github.com/databus23/helm-diff/issues/503
+	f.BoolVar(&diff.insecureSkipTLSVerify, "helm-insecure-skip-tls-verify", false, "skip tls certificate checks for the chart download")
 	f.BoolVar(&diff.normalizeManifests, "normalize-manifests", false, "normalize manifests before running diff to exclude style differences from the output")
 
 	AddDiffOptions(f, &diff.Options)


yxxhero:
@mumoshu

yxxhero:
WDYT?

cmd/upgrade.go
```diff
@@ -253,7 +253,8 @@ func newChartCommand() *cobra.Command {
 	f.BoolVar(&diff.enableDNS, "enable-dns", false, "enable DNS lookups when rendering templates")
 	f.StringVar(&diff.postRenderer, "post-renderer", "", "the path to an executable to be used for post rendering. If it exists in $PATH, the binary will be used, otherwise it will try to look for the executable at the given path")
 	f.StringArrayVar(&diff.postRendererArgs, "post-renderer-args", []string{}, "an argument to the post-renderer (can specify multiple)")
-	f.BoolVar(&diff.insecureSkipTLSVerify, "insecure-skip-tls-verify", false, "skip tls certificate checks for the chart download")
+	// see https://github.com/databus23/helm-diff/issues/503
+	f.BoolVar(&diff.insecureSkipTLSVerify, "helm-insecure-skip-tls-verify", false, "skip tls certificate checks for the chart download")
```
mumoshu, line 257: Thanks! Just looking into it; How is this supposed to fix the mentioned issue?

Is it that `helm` somehow thinks `--insecure-skip-tls-verify` is for helm, not for the helm plugin(=helm-diff in this case)?

yxxhero: see https://github.com/helm/helm/blob/15d116b8548f59959fe9f2fafdaf7c6de8db3ea0/cmd/helm/load_plugins.go#L154:  
```
kvargs := []string{"--kube-context", "--namespace", "-n", "--kubeconfig", "--kube-apiserver", "--kube-token", "--kube-as-user", "--kube-as-group", "--kube-ca-file", "--registry-config", "--repository-cache", "--repository-config", "--insecure-skip-tls-verify", "--tls-server-name"}
```
`--insecure-skip-tls-verify` is so special, helm will not pass it into plugin. it's helm's rule.

yxxhero: @mumoshu

mumoshu: Thanks! I believe I understand that. My questions stems from that; I don't understand why that helm behavior result in `unknown flag: --insecure-skip-tls-verify` we are facing.

If helm doesn't process it AND send it to helm-diff, neither helm nor helm-diff should result in `unknown flag: --insecure-skip-tls-verify`.

Given the helm rule you mentioned, how does it result in this error, and why renaming this flag in helm-diff fixes that?

yxxhero: The key reason is helm will set it  as global flag, but helm has no the global flag.  @mumoshu

mumoshu: @yxxhero Thanks again! Could you confirm this; Is it a helm bug that helm thinks `--insecure-skip-tls-verify` is "known" to helm in preprocessing, then helm fails by thinking it is actually "unknown" to helm when invoking a plugin? Helm messed up the plugin invocation path?

yxxhero: @mumoshu yeah. I get the result by reviewing the source code. I don't know why helm do this.

yxxhero: @mumoshu WDYT?

yxxhero:
see: https://github.com/helm/helm/pull/12856

yxxhero:
closed: https://github.com/helm/helm/pull/12856

PR #574, opened 2024-03-10
Add: if the default value is not specified, looks for values.yaml in …

monegim:
In helm, if the value file is not specified, `values.yaml` in the chart dir is considered as the default value.  This is not the case in the helm diff.


yxxhero:
@monegim please add some unitests.

Issue #246, opened 2020-11-12
diff does not honour "helm.sh/resource-policy: keep" 


timjohncurtis:
We have a deployment which used an older version of helm, which defines the namespace as a YAML object in the helm chart. We want to upgrade to removing this as is the defined method in helm 3, so have added `helm.sh/resource-policy: keep` as an annotation, however when I run `helm diff` on the chart, it shows it will be removed. This is intended functionality as the YAML no longer exists, but could a note be added to the output of the diff noting to the user that this is annotated so it will not be deleted?

chudyandrej:
I have the same problem.

ncronquist:
I also just ran into this. I know it may be kind of an edge case, but it might be worth re-opening.

PR #582, opened 2024-03-19
Add support for resource-policy: keep (#246)

phooijenga:
The annotation `helm.sh/resource-policy: keep` instructs Helm to skip deleting this resource when a Helm operation would result in its deletion. As the resource is not actually deleted, exclude it from the generated diff.


yxxhero:
@mumoshu @databus23 make sense.

yxxhero (approved):

Issue #474, opened 2023-06-20
Support the new set-literal argument


plnordquist:
Helm 3.12.0 seems to have introduced a set-literal argument to allow passing a literal value instead needing to escape strings with set or set-string.  The commit that introduced this feature is at https://github.com/helm/helm/commit/0833318 and the PR is at https://github.com/helm/helm/pull/9182.  The Helm Diff plugin should support this argument.

interone-ms:
Relevant for my usecase as well, I'd be grateful if this could be implemented!

interone-ms:
@plnordquist I had an attempt at this myself, if you're still interested you can fetch my fork's master branch and try it out. For me, it worked.

PR #590, opened 2024-04-15
Implement support for pass-through of --set-literal flag/values

interone-ms:
Hi all,

this PR enables support for `--set-literal`, see https://github.com/helm/helm/pull/9182 and issue #474 here.


interone-ms:
Tested locally on macOS and Linux. Don't have access to a Windows or FreeBSD machine, but there should be nothing in the PR that could cause issues.

PR #591, opened 2024-04-15
deps: use json-patch/v5 instead of incompatible module

mmorel-35:
Use module instead of incompatible version


yxxhero:
@mmorel-35 pls fix conflict.

yxxhero:
@mmorel-35 thanks so much.

Issue #605, opened 2024-04-30
Don't output help on `helm plugin install`


z0rc:
I'm installing helm-diff as part of CI pipeline. Every time plugin installation outputs help page.

I found this unwelcome, as in CI pipeline context this output is effectively a noise. Also it isn't possible to properly silence it. I can do `helm plugin install ... >/dev/null`, but this will also omit output from helm itself, which is actually valuable.

yxxhero:
PR is welcome

z0rc:
@yxxhero I can just remove https://github.com/databus23/helm-diff/blob/master/install-binary.sh#L146-L152 and https://github.com/databus23/helm-diff/blob/master/install-binary.sh#L166, right? Or do you expect something else?

yxxhero:
@mumoshu @databus23  WDYT?

yxxhero:
@z0rc PR is welcome.

PR #621, opened 2024-05-27
Don't output help on install

z0rc:
This isn't a real test, and it produces noise on each install, which isn't desired in case of running under CI environment.

Fixes https://github.com/databus23/helm-diff/issues/605.

### DIFF
--- a/install-binary.sh
+++ b/install-binary.sh
@@ -143,14 +143,6 @@ exit_trap() {
   exit $result
 }
 
-# testVersion tests the installed client to make sure it is working.
-testVersion() {
-  set +e
-  echo "$PROJECT_NAME installed into $HELM_PLUGIN_DIR/$PROJECT_NAME"
-  "${HELM_PLUGIN_DIR}/bin/diff" -h
-  set -e
-}
-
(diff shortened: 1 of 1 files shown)


yxxhero:
@z0rc how do we make sure that helm-diff can work well? maybe we can use `> /dev/null` to drop the output of `diff -h`?

z0rc:
@yxxhero by testing it during development/release process. I don't think testing after install is helpful for end user. Also I didn't found any other plugin that does this kind of testing nowadays. There is https://github.com/technosophos/helm-template from which `install-binary.sh` was [copied](https://github.com/databus23/helm-diff/commit/806a6e2c3d4bd93f50e6d6bdb7f1aea7a2bb1890#diff-84b911b5c187acb3fa01e487f0e0fda54285d6d67d5d630993bab014e4d8c942R3), but it's unmaintained.

yxxhero:
ok. make sense.

yxxhero:
@z0rc have a try.

z0rc:
Works as expected. Thanks.

PR #742, opened 2025-02-19
Add support for --take-ownership parameter

### LINKED ISSUE #731: support `--take-ownership` option like `helm install` and `helm upgrade` commands
# Background
Helm install and upgrade actions got support for `--take-ownership` option last week with [Helm v3.17.0 release](https://github.com/helm/helm/releases/tag/v3.17.0) ([PR](https://github.com/helm/helm/pull/13439)), that is required when importing existing k8s resources into Helm release (see reasons below).

# Feature request
Add support for `--take-ownership` option also for `helm diff` command provided by this plugin, so it would be possible to work around the failure from following command (for example when renaming Helm release):
* `helm diff upgrade --install $helmReleaseNewName ...`:
> Error: rendered manifests contain a resource that already exists. Unable to continue with install: `$k8sResourceKind "$k8sResourceName" in namespace "$k8sResourceNamespace" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "$helmReleaseNewName": current value is "$helmReleaseOldName"

by adding `--take-ownership` flag like this:
* `helm diff upgrade --install --take-ownership $helmReleaseNewName ...`
> Error: unknown flag: --take-ownership

# Alternatives discarded for working around similar solutions:
There are several alternative approaches that one might assume should work well, but they all have surprising and severe disadvantages (unless destroying k8s resources is acceptable):
* adding `meta.helm.sh/release-name` annotation for existing resources doesn't destroy them, but Helm starts ignoring them (won't change the resource based on Helm values and chart template)
* other options (like uninstalling old releas or installing new release with `--force` flag) are destroy the existing resources, that can cause incidents, for example:
  * application downtime due to deleting k8s Deployment
  * loosing access to cloud resources related to k8s resource managing to cloud provider resources

guillomep:
I trying to add support of the new `--take-ownership` argument that appears in Helm 3.17.

This works by trying to match manifests of the release with existing ones on the cluster. If the release mismatch, then the changing of ownership is displayed in the report.

I have not a lot experience in Go, nor on this project.

Fix #731

### DIFF
--- a/manifest/util.go
+++ b/manifest/util.go
@@ -1,0 +1,40 @@
+package manifest
+
+import (
+	"fmt"
+
+	jsoniter "github.com/json-iterator/go"
+)
+
+func deleteStatusAndTidyMetadata(obj []byte) (map[string]interface{}, error) {
+	var objectMap map[string]interface{}
+	err := jsoniter.Unmarshal(obj, &objectMap)
+	if err != nil {
+		return nil, fmt.Errorf("could not unmarshal byte sequence: %w", err)
+	}
+
+	delete(objectMap, "status")
+
+	metadata := objectMap["metadata"].(map[string]interface{})
+
+	delete(metadata, "managedFields")
+	delete(metadata, "generation")
+	delete(metadata, "creationTimestamp")
+	delete(metadata, "resourceVersion")
+	delete(metadata, "uid")
+
+	// See the below for the goal of this metadata tidy logic.
+	// https://github.com/databus23/helm-diff/issues/326#issuecomment-1008253274
+	if a := metadata["annotations"]; a != nil {
+		annotations := a.(map[string]interface{})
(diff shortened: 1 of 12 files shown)


yxxhero:
@guillomep nend more tests.

guillomep:
I added some tests.

manifest/util.go
yxxhero, line 1: add tests for this new files.

guillomep: Done

guillomep:
Do I have any remaining job to do ?

yxxhero:
@guillomep no more. latest question: did you test it locally?

guillomep:
Hello, yes I have tested it it looks like this

image

seboudry:
Hi!
Thanks for this new flag!
Usage is missing in readme :wink:

PR #762, opened 2025-04-03
Allow diff to include crds

obirhuppertz:
## Goal
This PR allows usage of `helm template` flag `--include-crds` during a `helm diff upgrade`.

### Before
When using `helmfile` with `helm diff upgrade` command you have to use post-renderer to get the fully rendered manifest and allow rendering of all resources including CRDs that are deployed via a sub-chart i.e. kyverno.

### Use Case
In order to be able to move a way from post-renderer and use `kustomization` builtin support of `helmfile` without breaking the manifest that you can get via `helm get manifest -n namespace release` it is needed to allow rendering the included crds if the chart decided to render these via a dependency.

It also makes `three-way-merge` more usefull as you now also get the diff for included crd changes.

### After
Merging this PR into an upcoming release would allow anyone to be able to catch changes to included crd changes during a `helm diff upgrade`.

### Breaking Changes?
None to be expected, the default is set to `false` which reflects the current behavior. You have to pass `--include-crds` excplitictly to also render included crds when calling template before diff when using three-way-merge
We used helm diff upgrade command in conjunction with helmfile with kustomize postrenderer.

### Follow up
`helmfile` maintainers could add this flag to `helmfile diff` command in future versions. Using just the patched version would need you to add `- "--include-crds"` to `helmDefaults.diffArgs` array.

(at)Others who read this - please upvote if you find this PR useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ Flags:
       --skip-schema-validation           disables rendered templates validation against the Kubernetes OpenAPI Schema
   -D, --find-renames float32             Enable rename detection if set to any value greater than 0. If specified, the value denotes the maximum fraction of changed content as lines added + removed compared to total lines in a diff for considering it a rename. Only objects of the same Kind are attempted to be matched
   -h, --help                             help for diff
+      --include-crds                     include CRDs in the diffing
       --include-tests                    enable the diffing of the helm test hooks
       --install                          enables diffing of releases that are not yet deployed via Helm (equivalent to --allow-unreleased, added to match "helm upgrade --install" command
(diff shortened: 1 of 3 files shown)


AMOKANDY:
Fine!. But please add the new option to “readme.md”.

yxxhero:
@obirhuppertz did you test this feature?

obirhuppertz:
@yxxhero yes, I did also made sure its non breaking by setting default to false i.e. not include the crds as it was before.

yxxhero:
@obirhuppertz done. could add feature PR for helmfile?

PR #881, opened 2025-11-18
fix: Use correct syntax for windows

stevehipwell:
This PR fixes the Windows support by using the correct syntax for environment variables.

### DIFF
--- a/plugin.yaml
+++ b/plugin.yaml
@@ -8,7 +8,10 @@ useTunnel: true
 platformCommand:
   - command: ${HELM_PLUGIN_DIR}/bin/diff
   - os: windows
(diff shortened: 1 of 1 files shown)


sushant-dev01:
You see platformHooks in error now because I downloaded a different version of helm-diff from github releases. The latest one says 
`failed to load plugins: failed to load plugin at "C:\Users\Ksush\AppData\Roaming\helm\plugins\helm-diff\plugin.yaml": error unmarshaling JSON: while decoding JSON: json: unknown field "args"`

stevehipwell:
@sushant-dev01 you need to be on at least Helm v3.18 for this to work.

PR #923, opened 2026-02-01
fix: detect manual changes on CRDs and CRs with three-way-merge

### LINKED ISSUE #917: helm diff --three-way-merge does not detect manual changes on Custom Resources (CRDs)
We are seeing inconsistent behavior with helm diff upgrade `--three-way-merge` when detecting manual changes made directly in the Kubernetes cluster, specifically for Custom Resources (CRDs).

For built-in Kubernetes resources (e.g. Deployment), helm diff `--three-way-merge` correctly detects live changes made via kubectl edit/patch.
However, for CRDs (example below uses KEDA ScaledObject), manual changes in the cluster are not detected, even when the field is explicitly defined in the Helm chart.

What works ✅
For built-in resources (e.g. Deployment):

1.Deploy via Helm
2.Manually edit:
`kubectl edit deployment my-app`
 change resources.limits.memory value

Run:
`helm diff upgrade <release> <chart> --three-way-merge --dry-run`

➡️ The manual change is detected and shown in the diff.

What does NOT work ❌
For CRDs (example: KEDA ScaledObject):

1. Deploy via Helm with spec.maxReplicaCount defined in the chart
2. Manually edit the ScaledObject:
```
kubectl patch scaledobject my-so \
  -n <ns> \
  --type=merge \
  -p '{"spec":{"maxReplicaCount":30}}'
```

Run:
`helm diff upgrade <release> <chart> --three-way-merge --dry-run`

➡️ No diff is shown, even though the live value differs from the chart.

Questions:
1. Is this a known limitation of helm-diff with CRDs?
2. Is `--three-way-merge` expected not to detect live drift for CRDs?
3. Are there plans or recommendations for supporting live drift detection on CRDs?

yxxhero:
## Summary
Fixes #917

Previously, `helm diff --three-way-merge` would not detect manual changes made to Custom Resources (CRDs and CRs) in the cluster. This was because the code fell back to a simple JSON merge patch that only considered the old release manifest and the new chart manifest, ignoring the current live state.

For unstructured objects (CRDs, CRs), we now perform a proper three-way merge that respects manual changes made in the cluster:

1. Create a patch from old -> new (chart changes)
2. Apply this patch to current (live state with manual changes)
3. Create a patch from current -> merged result
4. Return that patch (which will be applied to current by the caller)

This ensures that manual changes to CRDs and CRs are properly detected and displayed in the diff, just like they are for built-in Kubernetes resources.

## Changes
- Modified `manifest/generate.go` to implement three-way merge for unstructured objects (CRDs, CRs)
- Added detailed comments explaining the three-way merge approach
- All existing tests pass

## Testing
- Manual testing with CRDs and CRs confirms that manual changes are now detected
- All unit tests pass

### DIFF
--- a/manifest/generate.go
+++ b/manifest/generate.go
@@ -215,7 +215,10 @@ func createPatch(originalObj, currentObj runtime.Object, target *resource.Info) 
 			// Step 3: Create patch from current -> merged (what to apply to current)
 			// This patch, when applied to current, will produce the merged result
 			patch, err := jsonpatch.CreateMergePatch(cleanedCurrentData, mergedData)
-			return patch, types.MergePatchType, err
+			if err != nil {
+				return nil, types.MergePatchType, fmt.Errorf("creating patch from current to merged: %w", err)
+			}
+			return patch, types.MergePatchType, nil
 		}
 
 		// Chart didn't change (old == new), but we need to detect if current diverges
@@ -227,7 +230,10 @@ func createPatch(originalObj, currentObj runtime.Object, target *resource.Info) 
 			return nil, types.MergePatchType, fmt.Errorf("building desired state: %w", err)
 		}
 		patch, err := jsonpatch.CreateMergePatch(cleanedCurrentData, desiredData)
-		return patch, types.MergePatchType, err
+		if err != nil {
+			return nil, types.MergePatchType, fmt.Errorf("creating patch from current to desired: %w", err)
+		}
+		return patch, types.MergePatchType, nil
 	}
 
 	patchMeta, err := strategicpatch.NewPatchMetaFromStruct(versionedObject)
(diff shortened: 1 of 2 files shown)


yxxhero:
@terabyte128 PTAL.

yxxhero:
> > > @terabyte128 PTAL.
> > 
> > 
> > I tried with your updated changes and it correctly produces a diff! However, when using it with `helmfile`, the changes aren't actually applied. This might be an issue with `helmfile` and not `helm-diff`, though. For reference, this is what I did:
> > 
> > * manually edit a CRD using `kubectl edit`
> > * run `helmfile apply --interactive`
> > * see that the diff showed correctly, accept the change
> > 
> > But viewing the CRD manually still shows the old data.
> 
> I was able to replicate this without Helmfile. Running:
> 
> ```
> helm diff upgrade --reuse-values $CHART $SOURCE --context 5
> ```
> 
> produces a diff, but
> 
> ```
> helm upgrade --reuse-values $CHART $SOURCE
> ```
> 
> has no effect.

@terabyte128 so it's helm issue?

PR #1021, opened 2026-06-16
feat: add support for helm v4 --server-side flag

### LINKED ISSUE #1020: Please add support for the helm4 upgrade flag " --server-side"
Please add support for the helm4 upgrade flag " --server-side":

     --server-side string                         must be "true", "false" or "auto". Object updates run in the server instead of the client ("auto" defaults the value from the previous chart release's method) (default "auto")

This is needed to explicitly force Server-Side Apply for releases that were originally installed with Helm 3 and are being managed with Helm 4, where relying on "auto" may not be sufficient.

yxxhero:
## Summary

Add support for the Helm v4 `--server-side` upgrade flag, as requested in #1020.

The flag accepts `"true"`, `"false"`, or `"auto"` (default `"auto"`), matching [`helm v4 upgrade`](https://helm.sh/docs/helm/helm_upgrade/) exactly.

## Changes

- **`cmd/upgrade.go`**: Added `--server-side` flag (`StringVar`, default `"auto"`) with input validation in `RunE`.
- **`cmd/helm.go`**: Forward the flag to helm when running Helm v4:
  - `helm upgrade --dry-run` (`HELM_DIFF_USE_UPGRADE_DRY_RUN=true`): forward all values including `"auto"`.
  - `helm template`: forward only `"true"`/`"false"` — template registers `--server-side` as a `bool` (default `true`) that rejects `"auto"`.
  - Helm v3: flag is accepted but not forwarded (does not exist).
- **`README.md`**: Documented the flag in alphabetical order in both command sections.
- **Tests**: Added `TestServerSideFlagValidation` and `TestGetServerSideFlags`.

## Notes

- The flag has no effect on manifest rendering in dry-run mode — both `helm template` and `helm upgrade --dry-run` bail out before the apply step where `ServerSideApply` is consulted. It is forwarded for semantic correctness and wrapper compatibility.
- Single `isHelmVersionGreaterThanEqual` call (no redundant subprocess invocations).

Closes #1020

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -122,9 +122,9 @@ jobs:
           - helm-version: v3.21.1
           - helm-version: v4.2.1
     steps:
-      - uses: engineerd/setup-kind@v0.6.2
+      - uses: helm/kind-action@v1.14.0
         with:
-          skipClusterLogsExport: true
+          cluster_name: kind
 
       - uses: actions/checkout@v6
 
--- a/README.md
+++ b/README.md
@@ -164,6 +164,7 @@ Flags:
       --reset-then-reuse-values                  reset the values to the ones built into the chart, apply the last release's values and merge in any new values. If '--reset-values' or '--reuse-values' is specified, this is ignored
       --reset-values                             reset the values to the ones built into the chart and merge in any new values
       --reuse-values                             reuse the last release's values and merge in any new values. If '--reset-values' is specified, this is ignored
+      --server-side string                       must be "true", "false" or "auto". Object updates run in the server instead of the client ("auto" defaults the value from the previous chart release's method) (default "auto")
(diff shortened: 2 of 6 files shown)


databus23 (commented):
Nice PR overall, the feature is correct and well-documented. One minor suggestion on test structure:

`TestGetServerSideFlags` tests a standalone `getServerSideFlags()` helper defined in the test file that mirrors the production logic in `(*diffCmd).template()`. If the two ever diverge, the test still passes. It's testing a reimplementation, not the real code.

A cleaner approach: extract the inline switch block in `template()` into a package-level function (e.g. `serverSideFlags(isHelmV4 bool, useUpgradeDryRun bool, serverSide string) []string`), then call it from both `template()` and the test. That way the test exercises the actual production path.

PR #1026, opened 2026-06-18
Support array of scalars in structured output

demolitionmode:
## Summary

We've run into an issue when using `--output structured` in that changes to lists of strings (eg. those in `nodeSelectorTerms.matchExpressions.values`) does not get rendered properly.

## Current behaviour

Running the below command results in failure.

> helm diff upgrade -n apps <release> <chart> --output structured --values ...

Warning raised:

> Warning: failed to build structured entry for apps, k8s-cron, CronJob (batch) (kind: CronJob, changeType: MODIFY): Invalid JSON Document

Resulting output for that resource:

```json
[
  {
    "name": "apps, k8s-cron, CronJob (batch)",
    "changeType": "MODIFY",
    "resourceStatus": {
      "oldExists": false,
      "newExists": false
    }
  },
  ...
]
```

## Suspected cause

It appears that whilst generating a structured diff, maps and arrays of maps are handled correctly, but arrays of strings are not; they seem to be passed to [createNodePatch ](https://github.com/databus23/helm-diff/blob/46564405f798ba6a011b37fedf22dcfefc30ce28/diff/structured.go#L314) regardless, which tries to marshal and unmarshal these strings (which causes an error during the unmarshal).

## Proposed fix

Inside [diffArrayNodes](https://github.com/databus23/helm-diff/blob/46564405f798ba6a011b37fedf22dcfefc30ce28/diff/structured.go#L249), we can check if the two values are strings, then append a change immediately without a need for any further processing.

## Notes

This PR's first commit will just be adding to the relevant test case, which should show the behaviour mentioned above (once the workflow runs). The proposed fix will be pushed in a subsequent commit

### DIFF
--- a/diff/diff_test.go
+++ b/diff/diff_test.go
@@ -602,6 +602,15 @@ spec:
       containers:
       - name: app
         image: demo:v1
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: node-type
+                operator: In
+                values:
+                - standard
 `
 	newManifest := `
 apiVersion: apps/v1
@@ -616,6 +625,15 @@ spec:
       containers:
       - name: app
         image: demo:v2
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: node-type
+                operator: In
+                values:
+                - dedicated
 `
 	oldIndex := manifest.Parse([]byte(oldManifest), "prod", true)
 	newIndex := manifest.Parse([]byte(newManifest), "prod", true)
--- a/diff/diff_test.go
+++ b/diff/diff_test.go
@@ -651,7 +651,7 @@ spec:
 	require.Equal(t, "Deployment", entry.Kind)
 	require.Equal(t, "prod", entry.Namespace)
 	require.Equal(t, "web", entry.Name)
-	require.Len(t, entry.Changes, 2)
+	require.Len(t, entry.Changes, 3)
 	replicasChange, ok := findChange(entry.Changes, "spec", "replicas")
 	require.True(t, ok)
 	require.InDelta(t, float64(2), replicasChange.OldValue, 0.001)
@@ -661,6 +661,11 @@ spec:
 	require.True(t, ok)
 	require.Equal(t, "demo:v1", imageChange.OldValue)
 	require.Equal(t, "demo:v2", imageChange.NewValue)
+
+	affinityChange, ok := findChange(entry.Changes, "spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[0].values", "0")
+	require.True(t, ok)
+	require.Equal(t, "standard", affinityChange.OldValue)
+	require.Equal(t, "dedicated", affinityChange.NewValue)
 }
 
 func TestStructuredOutputAddAndRemove(t *testing.T) {
--- a/diff/diff_test.go
+++ b/diff/diff_test.go
@@ -613,8 +613,12 @@ spec:
                 - standard
       priorities:
       - 1
-      enabled:
+      booleanArrayField:
       - true
+      multiTypeField:
+      - 3
+      - true
+      - "string"
 `
 	newManifest := `
 apiVersion: apps/v1
@@ -640,8 +644,12 @@ spec:
                 - dedicated
       priorities:
       - 2
-      enabled:
+      booleanArrayField:
       - false
+      multiTypeField:
+      - false
+      - "new-string"
+      - 2
 `
 	oldIndex := manifest.Parse([]byte(oldManifest), "prod", true)
 	newIndex := manifest.Parse([]byte(newManifest), "prod", true)
@@ -659,7 +667,7 @@ spec:
 	require.Equal(t, "Deployment", entry.Kind)
 	require.Equal(t, "prod", entry.Namespace)
 	require.Equal(t, "web", entry.Name)
-	require.Len(t, entry.Changes, 5)
+	require.Len(t, entry.Changes, 8)
 	replicasChange, ok := findChange(entry.Changes, "spec", "replicas")
 	require.True(t, ok)
 	require.InDelta(t, float64(2), replicasChange.OldValue, 0.001)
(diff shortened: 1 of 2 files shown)


demolitionmode:
Thank you for approving the workflow, the error we're seeing is the same in the [test output](https://github.com/databus23/helm-diff/actions/runs/27788985022/job/82319061922#step:4:252) (`Warning: failed to build structured entry for prod, web, Deployment (apps) (kind: Deployment, changeType: MODIFY): Invalid JSON Document`).

The proposed fix has now been pushed for review.

---

## Results of local testing

### Using released version (3.15.10)

(I should've included this in the original description, apologies)

Unstructured output showing the expected changes

```bash
helm diff upgrade -n apps helm-diff-test chart
```

```diff
apps, k8s-cron, CronJob (batch) has changed:
  # Source: helm-diff-chart/templates/cron.yaml
  apiVersion: batch/v1
  kind: CronJob
  metadata:
    name: k8s-cron
    namespace: apps
    labels:
      app: helm-diff-test
  spec:
    schedule: "@yearly"
    jobTemplate:
      spec:
        template:
          metadata:
            labels:
              app: helm-diff-test
          spec:
            restartPolicy: "Never"
            containers:
            - name: k8s-cron
-             image: "hello-world"
+             image: "hello-world:1"
            affinity:
              nodeAffinity:
                requiredDuringSchedulingIgnoredDuringExecution:
                  nodeSelectorTerms:
                  - matchExpressions:
                    - key: node-type
                      operator: In
                      values:
-                     - standard
+                     - dedicated
apps, web, Deployment (apps) has changed:
  # Source: helm-diff-chart/templates/deployment.yaml
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: web
    namespace: apps
    labels:
      app: helm-diff-test
  spec:
    selector:
      matchLabels:
        app: helm-diff-test
    replicas: 1
    template:
      metadata:
        labels:
          app: helm-diff-test
      spec:
        containers:
        - name: app
          image: "hello-world"
        affinity:
          nodeAffinity:
            requiredDuringSchedulingIgnoredDuringExecution:
              nodeSelectorTerms:
              - matchExpressions:
                - key: node-type
                  operator: In
                  values:
-                 - standard
+                 - dedicated
```

Structured output failing:

```bash
helm diff upgrade -n apps helm-diff-test chart --output structured 
```

```
helm diff upgrade -n apps helm-diff-test chart --output structured
Warning: failed to build structured entry for apps, k8s-cron, CronJob (batch) (kind: CronJob, changeType: MODIFY): Invalid JSON Document
Warning: failed to build structured entry for apps, web, Deployment (apps) (kind: Deployment, changeType: MODIFY): Invalid JSON Document
[
  {
    "name": "apps, k8s-cron, CronJob (batch)",
    "changeType": "MODIFY",
    "resourceStatus": {
      "oldExists": false,
      "newExists": false
    }
  },
  {
    "name": "apps, web, Deployment (apps)",
    "changeType": "MODIFY",
    "resourceStatus": {
      "oldExists": false,
      "newExists": false
    }
  }
]
```

### Using patched binary

Structured output shows it's able to process arrays of strings properly now

```bash
HELM_NAMESPACE="apps" \
HELM_BIN="/opt/homebrew/bin/helm" \
  ./helm-diff upgrade helm-diff-test chart \
  --output structured
```

```json
[
  {
    "apiVersion": "batch/v1",
    "kind": "CronJob",
    "namespace": "apps",
    "name": "k8s-cron",
    "changeType": "MODIFY",
    "resourceStatus": {
      "oldExists": true,
      "newExists": true
    },
    "changes": [
      {
        "path": "spec.jobTemplate.spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[0].values",
        "field": "0",
        "change": "replace",
        "oldValue": "standard",
        "newValue": "dedicated"
      },
      {
        "path": "spec.jobTemplate.spec.template.spec.containers[0]",
        "field": "image",
        "change": "replace",
        "oldValue": "hello-world",
        "newValue": "hello-world:1"
      }
    ]
  },
  {
    "apiVersion": "apps/v1",
    "kind": "Deployment",
    "namespace": "apps",
    "name": "web",
    "changeType": "MODIFY",
    "resourceStatus": {
      "oldExists": true,
      "newExists": true
    },
    "changes": [
      {
        "path": "spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[0].values",
        "field": "0",
        "change": "replace",
        "oldValue": "standard",
        "newValue": "dedicated"
      }
    ]
  }
]
```

yxxhero (commented):
## Review — Approve with minor suggestions

Solid, minimal fix for a real bug, and the test reproduces the actual K8s case (`nodeSelectorTerms.matchExpressions.values`). Build/Test + Lint pass. Leaving a few suggestions below.

### The bug (context)
`diffArrayNodes` calls `createNodePatch` for every differing array element pair. `createNodePatch` delegates to `jsonpatch.CreateMergePatch`, which rejects non-object JSON documents — so two strings like `"standard"` vs `"dedicated"` yield `Invalid JSON Document` and the whole resource's structured entry is dropped.

### What's good
- Correct and minimal; mirrors the existing `subPatch == nil` replace path (structured.go:278-288), so output shape stays consistent.
- The `DeepEqual` check above already filters equal strings, so no redundant changes are emitted.
- Test asserts path, field, change type, old/new values and updates the change count 2 → 3.

### Suggestions

**1. Scope — other primitives have the same bug.** `createNodePatch` will also fail for arrays of numbers (`float64`) and booleans (`bool`) — e.g. a list of replica counts. The PR only fixes strings. Either widen the guard to all scalars, or add a comment noting the limitation so a future contributor isn't surprised:
```go
// createNodePatch only supports JSON objects; handle scalars directly.
```

**2. Flatten the nested `if` (style).** The current double-nested `if ok { if ok { ... } }` can be flattened by lifting the type assertions out of the condition:
```go
oldStr, oldIsStr := oldVal.(string)
newStr, newIsStr := newVal.(string)
if oldIsStr && newIsStr {
    path, field := splitTokens(next)
    *changes = append(*changes, FieldChange{
        Path:     path,
        Field:    field,
        Change:   "replace",
        OldValue: oldStr,
        NewValue: newStr,
    })
    continue
}
```
Behavior is identical — when `oldVal` is a string but `newVal` isn't, it still falls through to `createNodePatch`.

**3. Duplication (optional).** The `replace` FieldChange block now appears twice in this function (plus similar add/remove blocks). A small `appendReplaceChange(changes, next, old, new)` helper would reduce drift risk — optional given the existing style.

**4. Test coverage (nice-to-have).** Top-level string arrays and mixed-type arrays aren't covered. The add/remove paths for strings already work via the existing switch arms; just worth a test to lock it in.

yxxhero:
@demolitionmode thanks so much. please review my comments.

demolitionmode:
> @demolitionmode thanks so much. please review my comments.

Thanks, they're great suggestions, I'll work on getting them implemented

yxxhero:
@demolitionmode any updates? I will cut new release.

demolitionmode:
> @demolitionmode any updates? I will cut new release.

Sorry, I've not had the time yet to address these

demolitionmode:
@yxxhero apologies for the delay, I believe I've addressed the suggestions now (except for point 3, where I've kept to the existing style). I've added tests for mixed-type arrays, though I'm not sure which kubernetes objects accept these; it _is_ a nice-to-have in case any custom resources have such schema

PR #1030, opened 2026-07-10
install-binary.sh: Use $HELM_BIN in HELM_HOME assignment

zware:
When a command named 'helm' does not actually exist (for example when
Helm is invoked as 'helm4' or similar), an error like the following may
be printed even if installation then continues to apparently succeed.

   [/path/to]/helm/plugins/4/helm-diff/install-binary.sh: 19: helm: not found
   mkdir: cannot create directory '': No such file or directory


zware:
Thanks for the merge!

PR #1041, opened 2026-08-01
fix: strip OCI chart pull progress from rendered manifests (#1040)

### LINKED ISSUE #1040: [Bug] --three-way-merge fails when using OCI charts due to OCI pull output leaking into manifest buffer
# Description:

When using helm diff upgrade `--three-way-merge` with an OCI-hosted chart or subchart, Helm's OCI pull progress log is written to stdout and included in the rendered release manifest buffer.

When manifest/generate.go passes this buffer to kubeClient.Build(), Kubernetes fails to unmarshal the OCI log JSON because it lacks a Kind.

# Steps to Reproduce:

* Reference an OCI chart (e.g., public.ecr.aws/karpenter/karpenter) directly or as a subchart dependency.

* Run e.g. `helm diff upgrade --debug --normalize-manifests --install  --three-way-merge --version 1.9.0 --namespace karpenter karpenter oci://public.ecr.aws/karpenter/karpenter`

* Observe error:
```
llevel=DEBUG msg="executing plugin command" pluginName=diff command="/home/crohmann/.local/share/helm/plugins/helm-diff/bin/diff upgrade --normalize-manifests --install --three-way-merge --version 1.9.0 karpenter oci://public.ecr.aws/karpenter/karpenter"
Executing helm version
Executing helm get manifest karpenter --namespace karpenter
Executing helm get values karpenter --output yaml --all --namespace karpenter
Executing helm version
Executing helm version
Executing helm version
Executing helm version
Executing helm template karpenter oci://public.ecr.aws/karpenter/karpenter --version 1.9.0 --namespace karpenter --values /tmp/existing-values4265984746 --dry-run=server --is-upgrade
Error: unable to generate manifests: unable to build kubernetes objects from new release manifest: unable to decode "": Object 'Kind' is missing in '{"Digest":"sha256:8e3952caafd208cb888fbf97467cd04a4a024a3fba64c84af73039040cc6a371","Pulled":"public.ecr.aws/karpenter/karpenter:1.9.0"}'
level=DEBUG msg="plugin execution failed" pluginName=diff error="exit status 1" exitCode=1 stderr=""
```

* Manually calling `helm template karpenter oci://public.ecr.aws/karpenter/karpenter --version 1.9.0 --namespace karpenter --values /tmp/values.yaml --dry-run=server --is-upgrade`

shows the pull info that helm diff complains about right at the beginning:

```
Pulled: public.ecr.aws/karpenter/karpenter:1.9.0
Digest: sha256:8e3952caafd208cb888fbf97467cd04a4a024a3fba64c84af73039040cc6a371
level=WARN msg="This chart or one of its subcharts contains CRDs. Rendering may fail or contain inaccuracies."
---
# Source: karpenter/templates/poddisruptionbudget.yaml
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: karpenter
  namespace: karpenter

[...]
```

* switching to dry-run mode (not using helm template) via `HELM_DIFF_USE_UPGRADE_DRY_RUN=true helm diff upgrade --debug --normalize-manifests --install  --three-way-merge --version 1.9.0 --namespace karpenter karpenter oci://public.ecr.aws/karpenter/karpenter`also works:

```
level=DEBUG msg="executing plugin command" pluginName=diff command="/home/crohmann/.local/share/helm/plugins/helm-diff/bin/diff upgrade --normalize-manifests --install --three-way-merge --version 1.9.0 karpenter oci://public.ecr.aws/karpenter/karpenter"
Executing helm version
Executing helm get manifest karpenter --namespace karpenter
Executing helm get values karpenter --output yaml --all --namespace karpenter
Executing helm version
Executing helm version
Executing helm upgrade karpenter oci://public.ecr.aws/karpenter/karpenter --version 1.9.0 --namespace karpenter --values /tmp/existing-values4084653059 --server-side=auto --install --dry-run=server

```

# Expected Behavior / Idea for a bugfix

helm-diff should strip OCI pull output before passing the manifest buffer to kubeClient.Build() in (https://github.com/databus23/helm-diff/blob/01d88c2a943322246aca885a922f45770793b020/manifest/generate.go#L32-L35)

yxxhero:
## Problem

`helm diff upgrade --three-way-merge` (and `--take-ownership`) fails with OCI-hosted charts because Helm writes the chart pull progress to **stdout** before the rendered manifests:

```
Pulled: public.ecr.aws/karpenter/karpenter:1.9.0
Digest: sha256:8e3952caafd208cb888fbf97467cd04a4a024a3fba64c84af73039040cc6a371
---
# Source: ...
```

The `helm template` code path in helm-diff used a no-op filter, so these lines leaked into the manifest buffer and got parsed as a YAML document lacking a `Kind`, breaking `kubeclient.Build()`:

```
Error: unable to generate manifests: unable to build kubernetes objects from new release manifest:
unable to decode "": Object 'Kind' is missing in '{"Digest":"...","Pulled":"..."}'
```

This only affects the `helm template` path. The `HELM_DIFF_USE_UPGRADE_DRY_RUN=true` path works because `extractManifestFromHelmUpgradeDryRunOutput` only keeps the `MANIFEST:`/`HOOKS:` sections, discarding the leading progress output.

Fixes #1040.

## Fix

Strip Helm's OCI chart pull progress lines (`Pulled:`, `Digest:`, `Pulling:`) from the `helm template` output before it is returned, fixing the leak at the source. This benefits both `manifest.Generate()` (three-way-merge) and the `--take-ownership` `Build()` call, which both `kubeclient.Build()` the templated manifest.

### Why is the regex safe?

The lines are matched at the start of a line (column 0). Top-level Kubernetes manifest keys are `apiVersion`/`kind`/`metadata`/`spec` and are never `Pulled`/`Digest`/`Pulling`. Any homonymous keys nested inside a manifest (e.g. under `data:` or `metadata:`) are indented and therefore not matched.

## Verification

- `go build ./...`
- `go vet ./...`
- `golangci-lint run ./cmd/` → 0 issues
- `go test ./...` → all pass
- Added `TestStripOCIPullProgress` covering: progress prepended to manifests, all three line variants, no-op on clean output, indented keys preserved, progress-only input, and empty input.

## Checklist
- [x] The fix targets the root cause (the leaking `helm template` filter), so it covers three-way-merge **and** take-ownership paths.
- [x] No change to existing behavior for non-OCI charts.

### DIFF
--- a/cmd/helm.go
+++ b/cmd/helm.go
@@ -412,7 +412,7 @@ func (d *diffCmd) template(isUpgrade bool) ([]byte, error) {
 		subcmd = "template"
 
 		filter = func(s []byte) []byte {
-			return s
+			return stripOCIPullProgress(s)
 		}
 	}
 
@@ -492,6 +492,36 @@ func extractManifestFromHelmUpgradeDryRunOutput(s []byte, noHooks bool) []byte {
 	r = append(r, hooks...)
 
 	return r
+}
+
+// ociPullProgressRE matches Helm's OCI chart pull progress lines that Helm writes
+// to stdout before the rendered manifests when the chart, or one of its subcharts,
+// is pulled from an OCI registry.
+//
+// The lines reported in the wild are "Pulled: ..." and "Digest: ...";
+// "Pulling: ..." is matched defensively too.
+//
+// These lines are emitted at the start of a line and are not valid Kubernetes
+// manifests, so they are safe to strip. Top-level manifest keys are
+// apiVersion/kind/metadata/spec and never "Pulled", "Digest" or "Pulling";
+// any homonymous keys inside a manifest are indented and therefore not matched.
+//
+// See https://github.com/databus23/helm-diff/issues/1040
+var ociPullProgressRE = regexp.MustCompile(`(?m)^(?:Pulled|Digest|Pulling):[^\n]*\n?`)
+
+// stripOCIPullProgress removes Helm's OCI chart pull progress output that leaks
+// into the rendered manifest buffer when the chart (or a subchart) is pulled
+// from an OCI registry.
+//
+// Without this, the progress lines (e.g. "Pulled: ...", "Digest: ...") are
+// parsed as a YAML document lacking a Kind and break the downstream three-way
+// merge / kubeclient.Build():
+//
+//	unable to decode "": Object 'Kind' is missing in '{"Digest":"...","Pulled":"..."}'
+//
+// See https://github.com/databus23/helm-diff/issues/1040
+func stripOCIPullProgress(s []byte) []byte {
+	return ociPullProgressRE.ReplaceAll(s, []byte(""))
 }
 
 // serverSideFlags returns the --server-side flag(s) to forward to helm.
--- a/cmd/helm_test.go
+++ b/cmd/helm_test.go
@@ -8,6 +8,78 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
+func TestStripOCIPullProgress(t *testing.T) {
+	manifest := `---
+# Source: karpenter/templates/cm.yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: karpenter
+  namespace: karpenter
+data:
+  Pulled: preserved
+  Digest: preserved
+---
+# Source: karpenter/templates/deployment.yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: karpenter
+  namespace: karpenter
+`
+
+	cases := []struct {
+		name string
+		in   string
+		want string
+	}{
+		{
+			name: "strips OCI pull progress prepended to manifests",
+			in: "Pulled: public.ecr.aws/karpenter/karpenter:1.9.0\n" +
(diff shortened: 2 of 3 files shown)


sfbahr:
Hi @yxxhero, I'm experiencing a bug very close to what this PR fixed, but not quite.

I have OCI images with a "+" character in their semantic version, and that causes helm to output:
```
{registry}/{repo}:0.12.1_20130313144700 contains an underscore.

OCI artifact references (e.g. tags) do not support the plus sign (+). To support
storing semantic versions, Helm adopts the convention of changing plus (+) to
an underscore (_) in chart version tags when pushing to a registry and back to
a plus (+) when pulling from a registry.
```

AI noted to me that the output does go to stderr, but it gets combined with stdout within this plugin:
```
The failure is that helm-diff parses that stderr as manifest YAML. In
  cmd/helm.go:83 it uses cmd.CombinedOutput(), merging stderr into stdout, and
  elsewhere outputWithRichError folds exitError.Stderr into the returned bytes.
  Those bytes reach parseContent (manifest/parse.go:150-153), which does:
      log.Fatalf("YAML unmarshal error: %s\nCan't unmarshal %s", err, content)
  }
```

Edit: I run into this error message:
```
unable to build kubernetes objects from new release manifest:
  unable to decode "": json: cannot unmarshal string into Go value of type unstructured.detector
```

Issue #326, opened 2022-01-08
--three-way-merge produces useless diff on meta.helm.sh annotations


mumoshu:
Running `diff upgrade --three-way-merge` added via #304 on an unchanged release produces diffs on `meta.helm.sh/release-name` and `meta.helm.sh/release-namespace`. I guess this should be omitted from the diff as it's added automatically by `helm` AND running `helm upgrade` won't actually remove those annotations.

```
$ HELM_NAMESPACE=default HELM_BIN=helm372 ./helm-diff upgrade envoy1 stable/envoy --three-way-merge
default, envoy, Service (v1) has changed:
  apiVersion: v1
  kind: Service
  metadata:
-   annotations:
-     meta.helm.sh/release-name: envoy1
-     meta.helm.sh/release-namespace: default
    creationTimestamp: "2022-01-08T11:00:49Z"
    labels:
      app: envoy
      app.kubernetes.io/managed-by: Helm
      chart: envoy-1.9.4
      heritage: Helm
      release: envoy1
```

@luxurine @Temikus Have you experienced this? Would you agree that this should be fixed somehow?

TBBle:
It's not game-breaking, but it's a bit of a wart. I guess we could explicitly strip those two annotations from objects fetched from the cluster. Although then if someone has included them in the chart templates by-hand (unlikely, but not impossible, nor invalid), I guess they'd show up in the merge again, as 'drift'.

So stripping them from all three sets of resources (source, current, target) is probably feasible, and doesn't immediately jump out with any downsides.

Worth noting, the `app.kubernetes.io/managed-by: Helm` label is [handled the same way](https://github.com/helm/helm/blob/v3.7.2/pkg/action/validate.go#L112-L148), but because it predates the auto-added metadata, it's present in most people's Helm templates, which is why you're not seeing it in this test-case.

So handling all three of these Helm-added metadata items in one fix is a good way to observe and validate both cases.

mumoshu:
@TBBle Thank you so much for your feedback! I've enhanced helm-diff to ignore diff on meta.helm.sh annotations #331.

I didn't actually see any unnecessary diff on `app.kubernetes.io/managed-by: Helm` perhaps due to the chart I used for testing didn't have `app.kubernetes.io/managed-by` templates in it.

I'd definitely do the same care on the label as soon as I or anyone have seen the real instance of the problem.

TBBle:
You're very welcome.

[...]

PR #1044, opened 2026-08-12
feat: omit 'app.kubernetes.io/managed-by' label on three-way-diff

jkroepke:
Ref: https://github.com/databus23/helm-diff/issues/326#issuecomment-1008222454

> Worth noting, the app.kubernetes.io/managed-by: Helm label is [handled the same way](https://github.com/helm/helm/blob/v3.7.2/pkg/action/validate.go?rgh-link-date=2022-01-09T03%3A34%3A29.000Z#L112-L148), but because it predates the auto-added metadata, it's present in most people's Helm templates, which is why you're not seeing it in this test-case.

I'm using helm diff with three-way-diff in a upgrade --dry-mode and I observe changes on each resource which would remove `app.kubernetes.io/managed-by`.

### DIFF
--- a/manifest/util.go
+++ b/manifest/util.go
@@ -35,7 +35,7 @@ func deleteStatusAndTidyMetadata(obj []byte) (map[string]interface{}, error) {
 			delete(metadata, "annotations")
 		}
 	}
-	
+
 	if a := metadata["labels"]; a != nil {
 		labels := a.(map[string]interface{})
 		delete(labels, "app.kubernetes.io/managed-by")
--- a/manifest/util.go
+++ b/manifest/util.go
@@ -40,8 +40,8 @@ func deleteStatusAndTidyMetadata(obj []byte) (map[string]interface{}, error) {
 		labels := a.(map[string]interface{})
 		delete(labels, "app.kubernetes.io/managed-by")
 
-		if len(annotations) == 0 {
-			delete(labels, "annotations")
+		if len(labels) == 0 {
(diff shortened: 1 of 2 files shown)


yxxhero:
@jkroepke please not if in if.

yxxhero:
@jkroepke Thanks for the PR! I pushed a follow-up commit to avoid the nested `if` (no more "if in if").

I extracted a small helper so the two checks are sequential rather than nested:

```go
func pruneNestedMap(target map[string]interface{}, key string, fields ...string) {
	sub, ok := target[key].(map[string]interface{})
	if !ok {
		return
	}
	for _, field := range fields {
		delete(sub, field)
	}
	if len(sub) == 0 {
		delete(target, key)
	}
}
```

The call site is now flat and both the existing `annotations` tidy and the new `labels` tidy share it:

```go
pruneNestedMap(metadata, "annotations",
	"meta.helm.sh/release-name",
	"meta.helm.sh/release-namespace",
	"deployment.kubernetes.io/revision",
)
pruneNestedMap(metadata, "labels", "app.kubernetes.io/managed-by")
```

As a bonus the comma-ok assertion is a bit safer than the previous nil-check + type assertion. I also added test coverage for the label omission and the empty-labels cleanup path. PTAL.

jkroepke:
@yxxhero do you have a plan, when this gets released?

yxxhero:
@jkroepke soon.

PR #1048, opened 2026-08-19
Add external diff tool support

### LINKED ISSUE #638: Support unified format like kubectl diff for difftool compatibility (Meld)
Hello,

kubectl diff by default output in an `unified format like diffutils` does.
It also supports `KUBECTL_EXTERNAL_DIFF` environment variable which can be used to pass a gui tool like Meld.

Is it possible to add an unified format option to --output ?

Resources:

- https://kubernetes.io/docs/reference/kubectl/generated/kubectl_diff/
- https://www.gnu.org/software/diffutils/manual/diffutils.html#Unified-Format
- https://meldmerge.org/

Best regards.

eagafonov:
Add --diff-tool flag and HELM_DIFF_TOOL env var to render diffs with an external command.

Manifests are written to two temp files whose paths are appended as the last two arguments.

Secret redaction and line suppression stay in effect since manifests are reconstructed from the report entries.

Exit code 1 (differences found) is ignored; other failures report to stderr without aborting helm-diff.

The patch addresses https://github.com/databus23/helm-diff/issues/638

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ When a kind is suppressed via `--suppress`, `changesSuppressed` is set to `true`
 
 ### External diff tool
 
-Set `--diff-tool` to a command and helm-diff renders the diff with that command instead of its built-in renderers. It writes the old and the new manifests into two temporary files and appends their paths as the last two arguments:
+Set `--diff-tool` to a command and helm-diff renders the diff with that command instead of its built-in renderers. It writes the old and the new manifests into two temporary files (named `old.yaml` and `new.yaml` in a private temporary directory) and appends their paths as the last two arguments:
 
 ```shell
 # any tool that accepts two file paths works
@@ -240,12 +240,15 @@ export HELM_DIFF_TOOL="difft --language yaml"
 helm diff upgrade api ./charts/api
 ```
 
-`--diff-tool` takes precedence over `HELM_DIFF_TOOL`, and either one overrides `--output`. There is no default command: without one, the built-in `--output` renderer is used.
+`--diff-tool` takes precedence over `HELM_DIFF_TOOL`, and either one overrides `--output`. An explicit `--output` (or an explicitly empty `--diff-tool ""`) also wins over `HELM_DIFF_TOOL`, so scripts that parse a specific output format cannot be silently broken by a variable inherited from a shell profile: the environment variable only applies when neither flag is given. There is no default command: without one, the built-in `--output` renderer is used.
 
 Notes:
 
-- The command is executed directly, not through a shell, so pipes and shell expansion are not available. Wrap arguments containing spaces in quotes, for example `--diff-tool '"/opt/my tools/diff" -u'`. For anything more involved, point the flag at a wrapper script.
-- The manifests handed to the tool are the ones from the diff report, so `--suppress`, `--suppress-output-line-regex` and secret redaction still apply. Secrets are redacted unless `--show-secrets` is given, and suppressed kinds are replaced by a placeholder on both sides.
+- The command is executed directly, not through a shell, so pipes and shell expansion are not available. Wrap arguments containing spaces in quotes, for example `--diff-tool '"/opt/my tools/diff" -u'`; an unclosed quote is rejected with an error. For anything more involved, point the flag at a wrapper script.
+- Each resource in the temporary files is preceded by a `# Resource:`/`# Change:` header comment (change types: `ADD`, `REMOVE`, `MODIFY`, `OWNERSHIP`), because an external tool would otherwise have no way to show them.
+- The manifests handed to the tool are the ones from the diff report, so `--suppress`, `--suppress-output-line-regex` and secret redaction still apply. Secrets are redacted unless `--show-secrets` is given; suppressed kinds — and entries whose diff is empty after `--suppress-output-line-regex` — are replaced by a placeholder on both sides.
+- The command must block until it has finished reading the two files. GUI tools that return immediately (for example `code --diff`) may find the temporary directory already deleted before they display it; make them wait (for example `code --wait --diff`) or wrap them in a script that waits.
+- There is no timeout around the command: a tool that never exits keeps helm-diff running.
 - An exit code of `1` from the tool is treated as "differences found" and ignored. Other failures are reported on stderr without aborting helm-diff.
 - helm-diff's own exit code is unaffected by the tool: `--detailed-exitcode` still returns `2` based on the changes helm-diff detected.
 - `--context`/`-C` is not applied; use the equivalent option of the external tool (for example `diff -U3`).
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ export HELM_DIFF_TOOL="difft --language yaml"
 helm diff upgrade api ./charts/api
 ```
 
-`--diff-tool` takes precedence over `HELM_DIFF_TOOL`, and either one overrides `--output`. An explicit `--output` (or an explicitly empty `--diff-tool ""`) also wins over `HELM_DIFF_TOOL`, so scripts that parse a specific output format cannot be silently broken by a variable inherited from a shell profile: the environment variable only applies when neither flag is given. There is no default command: without one, the built-in `--output` renderer is used.
+`--diff-tool` overrides both `--output` and `HELM_DIFF_TOOL`. `HELM_DIFF_TOOL` applies only when neither `--output` nor `--diff-tool` is explicitly given, so scripts that parse a specific output format cannot be broken by a variable inherited from a shell profile; an explicitly empty `--diff-tool ""` disables the external tool. There is no default command: without one, the built-in `--output` renderer is used.
 
 Notes:
 
--- a/README.md
+++ b/README.md
@@ -146,6 +146,7 @@ Flags:
   -C, --context int                              output NUM lines of context around changes (default -1)
       --detailed-exitcode                        return a non-zero exit code when there are changes
       --devel                                    use development versions, too. Equivalent to version '>0.0.0-0'. If --version is set, this is ignored.
+      --diff-tool string                         command used to compare the manifests instead of the built-in --output renderers (can also be set via the env var HELM_DIFF_TOOL). The old and the new manifest file paths are appended as the last two arguments
       --disable-openapi-validation               disables rendered templates validation against the Kubernetes OpenAPI Schema
       --disable-validation                       disables rendered templates validation against the Kubernetes cluster you are currently pointing to. This is the same validation performed on an install
       --dry-run string[="client"]                --dry-run, --dry-run=client, or --dry-run=true disables cluster access and show diff as if it was install. Implies --install, --reset-values, and --disable-validation. --dry-run=server enables the cluster access with helm-get and the lookup template function.
@@ -220,6 +221,35 @@ helm diff upgrade prod api ./charts/api --output structured
 
 When a kind is suppressed via `--suppress`, `changesSuppressed` is set to `true` and field details are omitted. Nested metadata such as labels show the container path (`metadata.labels`) and expose the label key through the `field` property (for example `app.kubernetes.io/version`).
 
+### External diff tool
+
+Set `--diff-tool` to a command and helm-diff renders the diff with that command instead of its built-in renderers. It writes the old and the new manifests into two temporary files and appends their paths as the last two arguments:
+
+```shell
+# any tool that accepts two file paths works
+helm diff upgrade api ./charts/api --diff-tool "diff -u -N"
+helm diff upgrade api ./charts/api --diff-tool "difft --language yaml"
+helm diff upgrade api ./charts/api --diff-tool "git --no-pager diff --no-index --color"
+helm diff upgrade api ./charts/api --diff-tool "delta --side-by-side"
+```
+
+The command can also be set through the `HELM_DIFF_TOOL` environment variable, which is convenient in a shell profile:
+
+```shell
+export HELM_DIFF_TOOL="difft --language yaml"
+helm diff upgrade api ./charts/api
+```
+
+`--diff-tool` takes precedence over `HELM_DIFF_TOOL`, and either one overrides `--output`. There is no default command: without one, the built-in `--output` renderer is used.
+
+Notes:
+
+- The command is executed directly, not through a shell, so pipes and shell expansion are not available. Wrap arguments containing spaces in quotes, for example `--diff-tool '"/opt/my tools/diff" -u'`. For anything more involved, point the flag at a wrapper script.
+- The manifests handed to the tool are the ones from the diff report, so `--suppress`, `--suppress-output-line-regex` and secret redaction still apply. Secrets are redacted unless `--show-secrets` is given, and suppressed kinds are replaced by a placeholder on both sides.
+- An exit code of `1` from the tool is treated as "differences found" and ignored. Other failures are reported on stderr without aborting helm-diff.
+- helm-diff's own exit code is unaffected by the tool: `--detailed-exitcode` still returns `2` based on the changes helm-diff detected.
+- `--context`/`-C` is not applied; use the equivalent option of the external tool (for example `diff -U3`).
+
 ## Commands:
 
 ### local:
@@ -249,6 +279,7 @@ Flags:
   -a, --api-versions stringArray                 Kubernetes api versions used for Capabilities.APIVersions
   -C, --context int                              output NUM lines of context around changes (default -1)
       --detailed-exitcode                        return a non-zero exit code when there are changes
+      --diff-tool string                         command used to compare the manifests instead of the built-in --output renderers (can also be set via the env var HELM_DIFF_TOOL). The old and the new manifest file paths are appended as the last two arguments
       --enable-dns                               enable DNS lookups when rendering templates
   -D, --find-renames float32                     Enable rename detection if set to any value greater than 0. If specified, the value denotes the maximum fraction of changed content as lines added + removed compared to total lines in a diff for considering it a rename. Only objects of the same Kind are attempted to be matched
   -h, --help                                     help for local
@@ -329,6 +360,7 @@ Flags:
   -C, --context int                              output NUM lines of context around changes (default -1)
       --detailed-exitcode                        return a non-zero exit code when there are changes
       --devel                                    use development versions, too. Equivalent to version '>0.0.0-0'. If --version is set, this is ignored.
+      --diff-tool string                         command used to compare the manifests instead of the built-in --output renderers (can also be set via the env var HELM_DIFF_TOOL). The old and the new manifest file paths are appended as the last two arguments
       --disable-openapi-validation               disables rendered templates validation against the Kubernetes OpenAPI Schema
       --disable-validation                       disables rendered templates validation against the Kubernetes cluster you are currently pointing to. This is the same validation performed on an install
       --dry-run string[="client"]                --dry-run, --dry-run=client, or --dry-run=true disables cluster access and show diff as if it was install. Implies --install, --reset-values, and --disable-validation. --dry-run=server enables the cluster access with helm-get and the lookup template function.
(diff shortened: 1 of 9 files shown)


cmd/options.go
```diff
@@ -14,6 +14,7 @@ func AddDiffOptions(f *pflag.FlagSet, o *diff.Options) {
 	f.StringArrayVar(&o.SuppressedKinds, "suppress", []string{}, "allows suppression of the kinds listed in the diff output (can specify multiple, like '--suppress Deployment --suppress Service')")
 	f.IntVarP(&o.OutputContext, "context", "C", -1, "output NUM lines of context around changes")
 	f.StringVar(&o.OutputFormat, "output", "diff", "Possible values: diff, simple, template, json, structured, dyff. When set to \"template\", use the env var HELM_DIFF_TPL to specify the template.")
+	f.StringVar(&o.DiffToolCommand, "diff-tool", "", "command used to compare the manifests instead of the built-in --output renderers (can also be set via the env var HELM_DIFF_TOOL). The old and the new manifest file paths are appended as the last two arguments")
```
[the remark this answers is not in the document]
eagafonov: fixed

README.md
```diff
@@ -219,6 +220,35 @@ helm diff upgrade prod api ./charts/api --output structured
...
+### External diff tool
+
+Set `--diff-tool` to a command and helm-diff renders the diff with that command instead of its built-in renderers. It writes the old and the new manifests into two temporary files and appends their paths as the last two arguments:
+
+```shell
+# any tool that accepts two file paths works
+helm diff upgrade prod api ./charts/api --diff-tool "diff -u -N"
+helm diff upgrade prod api ./charts/api --diff-tool "difft --language yaml"
+helm diff upgrade prod api ./charts/api --diff-tool "git --no-pager diff --no-index --color"
+helm diff upgrade prod api ./charts/api --diff-tool "delta --side-by-side"
```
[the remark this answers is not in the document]
eagafonov: Appologies for the AI slop :facepalm: 
I had to review it more carefully.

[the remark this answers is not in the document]
eagafonov: fixed

diff/difftool.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// writeDiffToolSides reconstructs the old and the new manifests from the report
+// entries rather than from the raw manifests, which keeps secret redaction and line
+// suppression in effect for whatever the diff tool receives.
+func writeDiffToolSides(r *Report, oldPath, newPath string) error {
+	var current, next strings.Builder
+
+	for _, entry := range r.Entries {
+		header := "# Source: " + entry.Key + "\n"
```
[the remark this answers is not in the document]
eagafonov: fixed

yxxhero:
@eagafonov please fix ci issue.

eagafonov:
@yxxhero 

> please fix ci issue.

Fixed in a separate commit as a uniform refactor

Here is weird thing: my 'diff tool' patch hasn't introduced the issue, but somehow reviled it.
The latest golangci-lint v2.13.0 does not report it on my original code, but golangci-lint v2.12.2 (from Github CI) does report it. Anyway, there are separate refactor commit to extract all consts.

diff/diff.go
```diff
@@ -30,13 +30,38 @@ type Options struct {
...
+// output skips.
 func (o *Options) StructuredOutput() bool {
-	return o != nil && o.OutputFormat == "structured"
+	return o != nil && o.OutputFormat == outputFormatStructured && !o.DiffTool()
+}
+
+// DiffTool reports whether the diff is rendered by an external tool. Configuring a
+// command is the only way to ask for it, and it overrides the built-in outputs.
+func (o *Options) DiffTool() bool {
+	return o != nil && diffToolCommand(o.DiffToolCommand) != ""
```
yxxhero: **Main issue #1: env var silently overrides an explicit `--output`.**

Since the README recommends putting `export HELM_DIFF_TOOL=...` in the shell profile, anyone on that machine running a CI script with `helm diff ... --output json` (whose stdout is parsed by a downstream tool) will silently get free-form text from the external tool instead, breaking the script in a way that's hard to debug.

Suggestion: in `ProcessDiffOptions` (which has the FlagSet, so `f.Changed("output")` is available), let an explicitly-set `--output` win over the env var — or at least print a warning to stderr. The precedence chain *explicit flag > explicit env > default* is safer for unattended use.

diff/difftool.go
```diff
@@ -0,0 +1,174 @@
...
+	return filepath.Join(dir, "current.yaml"),
+		filepath.Join(dir, "new.yaml"),
+		func() { _ = os.RemoveAll(dir) },
+		nil
+}
+
+// writeDiffToolSides reconstructs the old and the new manifests from the report
+// entries rather than from the raw manifests, which keeps secret redaction and line
+// suppression in effect for whatever the diff tool receives.
+func writeDiffToolSides(r *Report, oldPath, newPath string) error {
```
yxxhero, line 153: **Main issue #2: `--suppress-output-line-regex` leaves no trace in tool mode.**

When every changed line of a MODIFY entry is filtered out, `doSuppress` flips the entry to `MODIFY_SUPPRESSED` with empty `Diffs` (diff.go:196-200). `writeDiffToolSides` then writes only the header on both sides → the external tool reports *no differences*, yet helm-diff still exits 2 (entries exist and `seenAnyChanges` is computed before suppression). The user gets the contradictory combination of "exit code 2 but empty diff". The built-in outputs handle this by printing "has changed, but diff is empty after suppression."

Note the `--suppress` (SuppressedKinds) path right below does this correctly — identical placeholder on both sides. Suggest doing the same for `MODIFY_SUPPRESSED` (and for an ADD whose lines were all filtered out), e.g. `# Changes suppressed by --suppress-output-line-regex`.

diff/difftool.go
```diff
@@ -0,0 +1,174 @@
...
+		fmt.Fprintf(os.Stderr, "Error: no diff tool configured\n")
+		return
+	}
+
+	oldFile, newFile, cleanup, err := createDiffToolFiles()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: unable to create temporary files for the diff tool: %v\n", err)
+		return
+	}
+	defer cleanup()
```
yxxhero, line 112: **Main issue #4: GUI tools that return immediately race with `defer cleanup()`.**

Tools like `code --diff` (VS Code returns immediately) or daemonized viewers will have the temp dir removed by `RemoveAll` right after `cmd.Run()` returns, leaving an empty diff panel. Since GUI diff viewers (Meld, VS Code — see #638) are the primary motivation for this feature, this deserves at least a README note: the command must block until it's done reading the files, otherwise wrap it in a script that waits.

diff/difftool.go
```diff
@@ -0,0 +1,174 @@
...
+
+func createDiffToolFiles() (oldFile, newFile string, cleanup func(), err error) {
+	// Stable basenames in a private directory: diff tools label their output
+	// with the file names, which randomized temp names would make unreadable.
+	dir, err := os.MkdirTemp("", "helm-diff-tool")
+	if err != nil {
+		return "", "", nil, err
+	}
+
+	return filepath.Join(dir, "current.yaml"),
```
yxxhero: Nit: `current.yaml` is returned as `oldFile` but doesn't pair with `new.yaml` — `old.yaml`/`new.yaml` would read better, since tools label their hunks/panels with these file names (that's the stated reason for stable basenames, and it works nicely).

Also, manifest `Content` already begins with its own `---\n# Source: <template path>` header, so the injected `# Source: <key>` slightly duplicates it for regular entries — only OWNERSHIP entries genuinely need the injected one. Not blocking, just something to consider.

diff/difftool.go
```diff
@@ -0,0 +1,174 @@
...
+			current.WriteRune(r)
+			started = true
+		}
+	}
+
+	if started {
+		args = append(args, current.String())
+	}
+
+	return args
```
yxxhero: Minor: an unclosed quote is silently swallowed — `splitDiffToolCommand("diff -u \"foo")` yields `["diff", "-u", "foo"]` with the quote just gone, which will confuse users. Consider returning an error from the tokenizer. No `\"` escaping either — acceptable given the README already points complex cases at a wrapper script.

cmd/options_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	return o
+}
+
+func TestProcessDiffOptionsSuppressSecrets(t *testing.T) {
+	o := processedOptions(t, "--suppress-secrets")
+	require.Contains(t, o.SuppressedKinds, "Secret")
+}
+
+func TestAddDiffOptionsHasNoExternalOutputFormat(t *testing.T) {
```
yxxhero: This test asserts the `--output` usage string does *not* contain the word "external" — it's brittle (a future, perfectly reasonable mention of `--diff-tool` in that usage text would break it) and doesn't verify any actual behavior. I'd drop it or turn it into a positive assertion.

README.md
```diff
@@ -220,6 +221,35 @@ helm diff upgrade prod api ./charts/api --output structured
...
+helm diff upgrade api ./charts/api
+```
+
+`--diff-tool` takes precedence over `HELM_DIFF_TOOL`, and either one overrides `--output`. There is no default command: without one, the built-in `--output` renderer is used.
+
+Notes:
+
+- The command is executed directly, not through a shell, so pipes and shell expansion are not available. Wrap arguments containing spaces in quotes, for example `--diff-tool '"/opt/my tools/diff" -u'`. For anything more involved, point the flag at a wrapper script.
+- The manifests handed to the tool are the ones from the diff report, so `--suppress`, `--suppress-output-line-regex` and secret redaction still apply. Secrets are redacted unless `--show-secrets` is given, and suppressed kinds are replaced by a placeholder on both sides.
+- An exit code of `1` from the tool is treated as "differences found" and ignored. Other failures are reported on stderr without aborting helm-diff.
```
yxxhero, line 252: Two suggestions for this Notes list, related to the main review points:

1. Note that ADD/REMOVE/OWNERSHIP change-type annotations from the built-in output are not visible when using an external tool — the tool only sees file content. A cheap improvement on the code side would be writing the change type into the header comment (e.g. `# Change: ADD`), which diff tools render as context lines.
2. Note that the command must block until it has finished reading the files — GUI tools that return immediately (e.g. `code --diff`) will see the temp files deleted before they display them.

yxxhero (commented):
Thanks for this PR — overall it's a well-executed feature and the code quality is above the bar for this repo. The core design decision — reconstructing the two file sides from the report entries rather than writing the raw manifests — means secret redaction, `--suppress`, `--suppress-output-line-regex`, `--strip-trailing-cr` and rename detection all keep working with zero duplicated filtering logic.

Also appreciated:

- Overriding the output printer at the report layer (instead of if/else in `cmd`), and correctly carrying `diffToolCommand` through `doSuppress` when the report is rebuilt — an easy detail to miss.
- Security posture: no shell involved (manifest content can never reach argv), 0700 private dir + 0600 files + `RemoveAll` cleanup — notably stricter than the existing `printDyffReport`, which ignores all temp-file errors.
- Exit-code semantics: the tool's `1` ("differences found") is ignored while helm-diff's own exit code (incl. `--detailed-exitcode` = 2) stays driven by the report — and it's documented.
- Thorough README (precedence, no default command, `--context` not applied, no shell features → wrapper script) and good tests, including all six `--output` formats being overridden.

I'd be happy to approve once the points below are discussed — **#1 and #2 look worth fixing before merge**; #3/#4 at minimum deserve a README note.

### Main issues

1. **`HELM_DIFF_TOOL` silently overrides an explicit `--output`** — a footgun for CI consumers (inline on `diff/diff.go`).
2. **`--suppress-output-line-regex` leaves no trace in tool mode** — user sees exit code 2 with an empty diff (inline on `diff/difftool.go`).
3. **ADD/REMOVE/OWNERSHIP semantics are invisible to the tool user** — suggest writing the change type into the header comment.
4. **GUI tools that return immediately race with `defer cleanup()`** — and GUI tools are the primary use case from #638.

### Minor

- `current.yaml` pairs with `new.yaml` but is returned as `oldFile` — naming is asymmetric; the manifest content already carries its own `# Source:` header, so the injected one slightly duplicates it.
- An unclosed quote in the command is silently dropped (`diff -u "foo` → `diff -u foo`).
- The constants refactor is incomplete: `outputFormatDiff` is missing and unknown formats still fall back silently (pre-existing behavior, not introduced here).
- `TestAddDiffOptionsHasNoExternalOutputFormat` asserts a help string *doesn't* contain "external" — brittle with no forward value.
- All non-trivial tests skip on `windows-latest`, which is in the CI matrix.
- No timeout around the external command — a hung tool hangs helm-diff (consistent with kubectl, fine, maybe worth a README line).
- Test gaps: exit-code-1 message content, end-to-end Secret redaction reaching the temp files, the `MODIFY_SUPPRESSED` integration path, and `--suppress` end-to-end (currently only covered at the `writeDiffToolSides` unit level).

diff/diff.go
yxxhero: **Main issue #1 fixed** in 9694a9c. `ProcessDiffOptions` now calls `o.IgnoreDiffToolEnvVar()` whenever `--output` or `--diff-tool` was explicitly given, implementing the *explicit flag > environment > default* chain: an explicit `--output` wins over `HELM_DIFF_TOOL`, and an explicitly empty `--diff-tool ""` becomes an explicit opt-out of the external tool. `Options.DiffTool()` only consults the env var when no explicit flag was given. Covered by new tests in `cmd/options_test.go`; README documents the precedence.

diff/difftool.go
yxxhero: **Main issue #2 fixed** in d667bdb. `writeDiffToolSides` now writes an identical `# Changes suppressed by --suppress-output-line-regex` placeholder on both sides whenever an entry ends up with empty diffs after suppression — which uniformly covers `MODIFY_SUPPRESSED` and ADD/REMOVE entries whose every changed line was filtered out (in `doSuppress` those keep their change type with empty `Diffs`). Same shape as the existing `--suppress` path: the tool reports no change for the entry, yet the user sees *why*. Covered end-to-end in `TestManifestsDiffToolOutput/honors_suppress-output-line-regex` plus a unit test for both entry kinds.

diff/difftool.go
yxxhero: Documented in 81765b4: the README notes now say the command must block until it has finished reading the two files, that GUI tools returning immediately (e.g. `code --diff`) may find the temp directory deleted before they display it, and suggest making them wait (`code --wait --diff`) or wrapping them in a script. I kept `cmd.Run()` blocking rather than adding wait/poll heuristics for daemonized viewers — a wrapper script remains the flexible answer.

diff/difftool.go
yxxhero: Renamed to `old.yaml`/`new.yaml` in d667bdb (and a test now asserts the basenames via the args the tool receives). On the header: the injected line now reads `# Resource: <key>` instead of `# Source:`, so it no longer collides with the manifest content's own `# Source: <template path>` line — the two carry different information (namespace/name/kind vs. template path), and `# Resource:` pairs naturally with the new `# Change:` line from main issue #3.

diff/difftool.go
yxxhero: Fixed in d667bdb: the tokenizer now returns an error for an unclosed quote and `printDiffToolReport` reports it on stderr instead of silently dropping it (`diff -u "foo` → error rather than `diff -u foo`). Escaping stays unsupported — the README keeps pointing complex cases at a wrapper script. Tests added for both quote styles.

cmd/options_test.go
yxxhero: Dropped in 9694a9c and replaced with positive assertions: the test now checks that the `--output` usage string lists the built-in formats (`diff, simple, template, json, structured, dyff`) and that the default value is `diff`.

README.md
yxxhero: Both notes added: (1) each resource is now preceded by a `# Resource: <key>` / `# Change: <ADD|REMOVE|MODIFY|OWNERSHIP>` header comment (d667bdb) — diff tools render them as context lines, so the change types are visible — and the README documents it (81765b4); (2) the README documents that the command must block until it has finished reading the files, with the `code --wait --diff` example. The timeout note you mentioned in the minors is in as well.

yxxhero:
@eagafonov I pushed three commits addressing every point from my review — thanks again for the well-executed feature. Inline replies are on the respective threads; summary:

**Main issues**
1. **`HELM_DIFF_TOOL` vs explicit `--output`** (9694a9c): explicit flag > environment > default — an explicit `--output` (or an explicitly empty `--diff-tool ""`) now wins over the env var, so CI scripts parsing `--output json` can't be broken by a shell profile.
2. **`--suppress-output-line-regex` leaving no trace** (d667bdb): fully suppressed entries now write an identical `# Changes suppressed by --suppress-output-line-regex` placeholder on both sides (covers `MODIFY_SUPPRESSED` and fully-filtered ADD/REMOVE), so no more exit 2 with an empty diff.
3. **Change types invisible** (d667bdb): each resource is preceded by `# Resource: <key>` / `# Change: <ADD|...>` header comments.
4. **GUI tools racing with cleanup** (81765b4): README note that the command must block, with the `code --wait --diff` example.

**Minors** — `old.yaml`/`new.yaml` basenames, unclosed-quote error instead of silent drop, `outputFormatDiff` constant + explicit fallback, brittle test replaced with positive assertions, POSIX-dependent tests rewritten with the `TestHelperProcess` pattern so they run on the Windows CI leg, and new coverage for exit-code-1 handling, failure reporting, end-to-end secret redaction / `--suppress` reaching the tool files, the `MODIFY_SUPPRESSED` path, and the basenames. `go test ./...`, `go vet` and golangci-lint v2.12.2 (the CI version) all pass locally.

Happy to approve once CI is green — nothing else outstanding on my side.

yxxhero:
Follow-up self-review pass (4dfa4bf), all polish — no behavior changes:

- **Centralized command resolution**: `DiffTool()` and the printer each derived the effective diff tool command independently (three `TrimSpace`/env reads across two functions); both now go through `Options.configuredDiffToolCommand()`, so the setup/print agreement is structural instead of an implicit invariant. Also refreshed the stale `DiffTool` doc comment.
- **Doc accuracy**: README now lists `MODIFY_SUPPRESSED` among the `# Change:` header values; the `writeDiffToolSides` comment now covers ADD/REMOVE/OWNERSHIP entries, which `doSuppress` does not rewrite.
- **Test hygiene**: removed a meaningless `require.NotPanics(t, func(){})` and a trivially-true assertion; documented the pipe-capacity constraint on `captureStderr`.

`go test -race`, `go vet` and golangci-lint v2.12.2 all pass.

yxxhero:
Third review pass (01c7eaf), one behavioral fix plus hardening:

- **Deferred printing re-read the environment**: the tool printer resolved the command via `HELM_DIFF_TOOL` at *print* time, so a `Report` obtained from `ManifestReport` and printed later rendered according to the environment at that later moment, and the "setup/print cannot drift apart" claim from the last pass was not structural. The Report now carries the command resolved once in `generateReport`; printing never reads the env. New test generates under `HELM_DIFF_TOOL` and prints with it unset.
- **README precedence paragraph contradicted itself** ("either one overrides `--output`" followed by "an explicit `--output` wins over `HELM_DIFF_TOOL`"); rewritten as one coherent `--diff-tool` > `HELM_DIFF_TOOL` (only when neither flag is explicit) > `--output` statement.
- **Windows test-helper hazard** (verified mechanically): `%q`-quoted Windows paths get their backslashes escaped, but the command tokenizer strips quotes *without* unescaping, so the re-exec'd test binary would receive a corrupted `C:\\Users\\...` path — and the Windows skips were removed two commits ago. The helper now quotes a forward-slash path (accepted by the OS on every platform, needs no escaping), with a round-trip regression test asserting the executable path survives the tokenizer.

Note for the maintainer: the CI workflows on this fork PR are in `action_required` state, so the Windows leg has not actually exercised these tests yet — the round-trip test is the local proxy for it. `go test -race ./...`, `go vet` and golangci-lint v2.12.2 all pass.

yxxhero:
Fourth review pass (0444caa). One correction to my previous note and one test cleanup:

- **Precision on the Windows remark**: the CI matrix's `windows-latest` legs only run `helm plugin install` — unit tests (`make test`) run exclusively on the ubuntu `build` job. So the portability work (helper-process tests, forward-slash quoting) is for local Windows runs and any future matrix change, not for current CI; the round-trip test remains the guard either way. My earlier "has not exercised these tests yet" phrasing suggested they eventually would — they won't, as things stand.
- Merged the last two subtests of `TestProcessDiffOptionsDiffTool`, which exercised the same `f.Changed("diff-tool")` branch with identical assertions; the surviving test also asserts the built-in renderer is selected.

Everything else came back clean on this pass (all five `ProcessDiffOptions` call sites order it before use; `Report.mode` is a pre-existing write-only field, out of scope here; no stale `current.yaml` references; `make test` runs `-race`, matching what was verified locally).

yxxhero:
@eagafonov please  review the latest changes.

yxxhero:
@eagafonov thanks so much, I will merge if all is ok.

yxxhero:
@eagafonov-nv

eagafonov:
> @eagafonov thanks so much, I will merge if all is ok.

Sh!p !t

PR #1056, opened 2026-08-27
feat: add --storage-namespace flag and HELM_DIFF_STORAGE_NAMESPACE env var

asouchang:
## Summary
In Kubernetes GitOps tooling (such as FluxCD's `HelmRelease` controller), workload resources are deployed to `targetNamespace` while Helm release storage records (Secrets/ConfigMaps `sh.helm.release.v1.<name>.v<n>`) are persisted in a separate `storageNamespace` (e.g. `flux-system`).

Previously, running `helm diff upgrade -n <targetNamespace>` attempted to fetch live release state from `<targetNamespace>`, failing with `release: not found`. Conversely, passing `-n <storageNamespace>` caused `helm template` to render manifests into the storage namespace instead of the target namespace.

This PR adds first-class support for `--storage-namespace` and the `HELM_DIFF_STORAGE_NAMESPACE` environment variable across `upgrade`, `revision`, and `rollback` commands.

## Changes
- Add `--storage-namespace` flag and `HELM_DIFF_STORAGE_NAMESPACE` env var across `upgrade`, `revision`, and `rollback`.
- Fetch live release manifests, hooks, values, and 3-way merge storage from `storageNamespace` while keeping `helm template` manifest rendering and unnamespaced resource parsing in `targetNamespace` (`-n` / `--namespace` / `HELM_NAMESPACE`).
- Fallback to `targetNamespace` when omitted for 100% backward compatibility.
- Add comprehensive unit tests covering flag parsing, env var resolution, and storage namespace fallback logic.
- Update documentation in `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,9 +416,11 @@ Flags:
   -D, --find-renames float32                     Enable rename detection if set to any value greater than 0. If specified, the value denotes the maximum fraction of changed content as lines added + removed compared to total lines in a diff for considering it a rename. Only objects of the same Kind are attempted to be matched
   -h, --help                                     help for release
       --include-tests                            enable the diffing of the helm test hooks
+      --kube-context string                      name of the kubeconfig context to use
       --normalize-manifests                      normalize manifests before running diff to exclude style differences from the output
       --output string                            Possible values: diff, simple, template, json, structured, dyff. When set to "template", use the env var HELM_DIFF_TPL to specify the template. (default "diff")
       --show-secrets                             do not redact secret values in the output
+      --show-secrets-decoded                     decode secret values in the output
       --strip-trailing-cr                        strip trailing carriage return on input
       --suppress stringArray                     allows suppression of the kinds listed in the diff output (can specify multiple, like '--suppress Deployment --suppress Service')
       --suppress-output-line-regex stringArray   a regex to suppress diff output lines that match
--- a/README.md
+++ b/README.md
@@ -156,8 +156,10 @@ Flags:
       --include-tests                            enable the diffing of the helm test hooks
       --insecure-skip-tls-verify                 skip tls certificate checks for the chart download
       --install                                  enables diffing of releases that are not yet deployed via Helm (equivalent to --allow-unreleased, added to match "helm upgrade --install" command
+      --kube-context string                      name of the kubeconfig context to use
       --kube-version string                      Kubernetes version used for Capabilities.KubeVersion
       --kubeconfig string                        This flag is ignored, to allow passing of this top level flag to helm
+  -n, --namespace string                         namespace to assume the release to be installed into. Defaults to the current kube config namespace.
       --no-color                                 remove colors from the output. If both --no-color and --color are unspecified, coloring enabled only when the stdout is a term and TERM is not "dumb"
       --no-hooks                                 disable diffing of hooks
       --normalize-manifests                      normalize manifests before running diff to exclude style differences from the output
@@ -188,7 +190,7 @@ Flags:
   -f, --values valueFiles                        specify values in a YAML file (can specify multiple) (default [])
       --version string                           specify the exact chart version to use. If this is not specified, the latest version is used
 
-Additional help topcis:
+Additional help topics:
   diff
 
 Use "diff [command] --help" for more information about a command.
@@ -329,6 +331,12 @@ HELM_DIFF_OUTPUT_CONTEXT=5 helm diff upgrade my-release datadog/datadog
   # This is equivalent to specifying the --storage-namespace flag.
   HELM_DIFF_STORAGE_NAMESPACE=flux-system helm diff upgrade -n prod-apps my-release datadog/datadog
 
+  # NOTE: The storage namespace separation is not supported in combination with
+  # HELM_DIFF_USE_UPGRADE_DRY_RUN=true, because rendering then goes through
+  # `helm upgrade --dry-run`, which resolves the release storage in the target
+  # namespace. If the release exists only in the storage namespace, keep the
+  # default `helm template` based rendering instead.
+
 Flags:
       --allow-unreleased                         enables diffing of releases that are not yet deployed via Helm
   -a, --api-versions stringArray                 Kubernetes api versions used for Capabilities.APIVersions
@@ -345,8 +353,10 @@ Flags:
       --include-tests                            enable the diffing of the helm test hooks
       --insecure-skip-tls-verify                 skip tls certificate checks for the chart download
       --install                                  enables diffing of releases that are not yet deployed via Helm (equivalent to --allow-unreleased, added to match "helm upgrade --install" command
+      --kube-context string                      name of the kubeconfig context to use
       --kube-version string                      Kubernetes version used for Capabilities.KubeVersion
       --kubeconfig string                        This flag is ignored, to allow passing of this top level flag to helm
+  -n, --namespace string                         namespace to assume the release to be installed into. Defaults to the current kube config namespace.
       --no-hooks                                 disable diffing of hooks
       --normalize-manifests                      normalize manifests before running diff to exclude style differences from the output
       --output string                            Possible values: diff, simple, template, json, structured, dyff. When set to "template", use the env var HELM_DIFF_TPL to specify the template. (default "diff")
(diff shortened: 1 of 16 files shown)


cmd/upgrade_test.go
```diff
@@ -233,3 +233,102 @@ func TestValidateRevision(t *testing.T) {
...
+	// When flag is not specified, RunE or initialization logic should pick up env var
+	cmd := newChartCommand()
+	_ = cmd.ParseFlags([]string{})
+
+	// Check resolution logic with env var
+	d := diffCmd{
+		namespace:        "my-target-ns",
+		storageNamespace: os.Getenv("HELM_DIFF_STORAGE_NAMESPACE"),
+	}
+	if d.getStorageNamespace() != "flux-system-env" {
```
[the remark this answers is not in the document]
asouchang: Added end-to-end command execution tests using fake Helm (capture_args mode) asserting helm get manifest targets the storage namespace while helm template targets the target namespace.

cmd/revision_test.go
```diff
@@ -0,0 +1,97 @@
...
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			os.Setenv("HELM_DIFF_STORAGE_NAMESPACE", tc.envVar)
+			r := revision{
+				namespace:        tc.namespace,
+				storageNamespace: tc.storageNamespace,
+			}
+			if r.storageNamespace == "" && tc.envVar != "" {
+				r.storageNamespace = os.Getenv("HELM_DIFF_STORAGE_NAMESPACE")
+			}
```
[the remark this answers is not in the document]
asouchang: Added end-to-end command execution tests with fake Helm asserting storage namespace flag and env var resolution.

cmd/rollback_test.go
```diff
@@ -0,0 +1,97 @@
...
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			os.Setenv("HELM_DIFF_STORAGE_NAMESPACE", tc.envVar)
+			r := rollback{
+				namespace:        tc.namespace,
+				storageNamespace: tc.storageNamespace,
+			}
+			if r.storageNamespace == "" && tc.envVar != "" {
+				r.storageNamespace = os.Getenv("HELM_DIFF_STORAGE_NAMESPACE")
+			}
```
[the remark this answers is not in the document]
asouchang: Added end-to-end command execution tests with fake Helm asserting storage namespace flag and env var resolution.

cmd/helpers.go
```diff
@@ -34,3 +35,22 @@ func outputWithRichError(cmd *exec.Cmd) ([]byte, error) {
...
+func resolveStorageNamespace(storageNamespace, namespace string) string {
+	if storageNamespace != "" {
+		return storageNamespace
+	}
+	return namespace
+}
+
+// resolveNamespaceFlags populates storageNamespace and namespace from their respective environment variables
+// (HELM_DIFF_STORAGE_NAMESPACE and HELM_NAMESPACE) if the flags were not explicitly set on the command line.
+func resolveNamespaceFlags(cmd *cobra.Command, storageNamespace, namespace *string) {
```
yxxhero: Minor: the `namespace` half of this helper is effectively a no-op in production — all three commands register `--namespace` with `os.Getenv("HELM_NAMESPACE")` as the flag default (evaluated at command construction), so `*namespace` already holds the env value by the time `RunE` runs, and re-reading the same env var can't change it.

It's harmless (and the tests rely on the re-read after `t.Setenv`), but having two mechanisms resolve the same thing invites drift. You could simplify by giving `--storage-namespace` the same treatment (`os.Getenv("HELM_DIFF_STORAGE_NAMESPACE")` as flag default) and dropping this helper entirely — an explicit `--storage-namespace=""` would still override the env since it sets the value to empty, which then falls back to the target namespace.

cmd/upgrade.go
```diff
@@ -332,7 +341,11 @@ func (d *diffCmd) runHelm3() error {
 	if d.threeWayMerge || d.takeOwnership {
 		actionConfig = new(action.Configuration)
 		localEnv := prepareEnvSettings(d.kubeContext)
-		if err := actionConfig.Init(localEnv.RESTClientGetter(), localEnv.Namespace(), os.Getenv("HELM_DRIVER")); err != nil {
+		storageNs := d.getStorageNamespace()
+		if storageNs == "" {
+			storageNs = localEnv.Namespace()
+		}
+		if err := actionConfig.Init(localEnv.RESTClientGetter(), storageNs, os.Getenv("HELM_DRIVER")); err != nil {
```
yxxhero, line 338: This looks right — `actionConfig.Init`'s namespace is where the release driver looks up the release record, so the storage namespace belongs here, and the `localEnv.Namespace()` fallback preserves the old behavior when neither namespace is set. `manifest.Generate` fetches live objects using each resource's own namespace from the manifest, so target-namespace workloads are unaffected. 👍

cmd/main_test.go
```diff
@@ -51,14 +51,31 @@ func TestMain(m *testing.M) {
...
+				} else {
+					fmt.Println(`version.BuildInfo{Version:"v3.18.0"}`)
 				}
+			} else {
+				fmt.Print(os.Getenv("HELM_DIFF_FAKE_OUTPUT"))
 			}
-			fmt.Print(os.Getenv("HELM_DIFF_FAKE_OUTPUT"))
 		default:
-			fmt.Print(os.Getenv("HELM_DIFF_FAKE_OUTPUT"))
+			if len(os.Args) > 1 && os.Args[1] == "version" {
```
yxxhero: Nit: this version-output block is now duplicated between the `capture_args` and `default` branches — consider extracting a small helper so future modes share it. Also, `HELM_DIFF_FAKE_VERSION_OUTPUT` is never exercised by any test; fine as an escape hatch, but a one-line test (or dropping the knob) would avoid carrying untested surface.

README.md
```diff
@@ -178,6 +178,7 @@ Flags:
       --show-secrets                             do not redact secret values in the output
       --show-secrets-decoded                     decode secret values in the output
       --skip-schema-validation                   skip validation of the rendered manifests against the Kubernetes OpenAPI schema
+      --storage-namespace string                 namespace where the helm release storage (Secret/ConfigMap) is located. Defaults to the target namespace (-n/--namespace)
```
yxxhero, line 183: Nit: the newly added `-n, --namespace` flag on upgrade/revision/rollback isn't listed in these flag tables (only the `--storage-namespace` rows were inserted). Consider regenerating the help blocks so the tables match the actual `--help` output.

README.md
```diff
@@ -323,6 +324,11 @@ Examples:
 # Read the flag usage below for more information on --context.
 HELM_DIFF_OUTPUT_CONTEXT=5 helm diff upgrade my-release datadog/datadog
 
+  # Set HELM_DIFF_STORAGE_NAMESPACE=flux-system to
+  # fetch release manifests/values/hooks from a storage namespace different from the target namespace.
```
yxxhero, line 330: Might be worth a caveat here: `HELM_DIFF_USE_UPGRADE_DRY_RUN=true` doesn't honor the storage/target separation — rendering then goes through `helm upgrade --dry-run --namespace <target>`, and helm resolves the release storage in the target namespace. With the release only present in the storage namespace, that invocation fails with "has no deployed releases" unless `--install`/`--allow-unreleased` is set (in which case it renders as a fresh install, which is still usable for diffing, just slightly different semantics for `.Release.Revision` etc.). A one-line note would help Flux users hitting this combination.

yxxhero (commented):
Thanks for the PR — this fills a real gap for FluxCD-style setups where the release storage (`sh.helm.release.v1.*` secrets) lives in a separate namespace from the workload target namespace.

I audited all namespace consumers in the three touched commands, and the separation looks correct:

- **Storage namespace (correctly switched):** `helm get manifest/hooks/values` (`getRelease`, `getHooks`, `writeExistingValues`) and `actionConfig.Init` for the three-way-merge / take-ownership path.
- **Target namespace (correctly kept):** `helm template` / `helm upgrade --dry-run` rendering, `manifest.Parse` namespace defaulting, and `checkOwnership` live-object lookups.

Backward compatibility is preserved via the `resolveStorageNamespace` fallback, and the error messages now report the namespace that was actually queried. The second commit also addresses the earlier review feedback (fake-helm `capture_args` e2e tests + shared helpers).

Verified locally on the PR branch: `go vet ./...`, `go test ./...` and `go test -race ./cmd/...` all pass.

The remaining comments are non-blocking suggestions. One scope note for a potential follow-up: `helm diff release` still resolves both releases through `HELM_NAMESPACE` / the `namespace/release` argument syntax only, so it has the same storage/target limitation this PR fixes for upgrade/revision/rollback — fine to leave out of this PR, but worth a follow-up issue for consistency.

